PR #25, opened 2025-09-04
add data_gen to /data and improve docs

dvweersel:


data/docker-compose.yml
```diff
@@ -50,5 +50,21 @@ services:
       POSTGRES_PASSWORD: ${PG_PASSWORD}
       POSTGRES_DB: ${PG_DB}
 
+  data_gen:
+    image: europe-west1-docker.pkg.dev/common-449414/common/zero-data:0.9.0
```
Boudewijn26, line 54: In hindsight, this would have been better to use `build: { context }` to actually depend on `zero-data` to keep it in sync

PR #28, opened 2025-09-09
Change amplifiers to use room ID instead of control id

dvweersel:
ZERO-658


zero-domestic-control/volumes/hass/home-assistant_v2.db
Boudewijn26, line 1: I'm pretty sure this guy is needed

dvweersel: Cloned the repo fresh, set everything up without issues. I believe Home assistant will just create the db if it doesn't exist.

dvweersel: Reverted removal. The `wal` and `shm` should be temp so keeping those out still.

PR #40, opened 2025-09-24
ZERO-795: scaffold THRS app

null-33:
Scherm­afbeelding 2025-09-24 om 14 01 29


zero-ui/src/components/modules/thrs/Breadcrumbs.vue
Boudewijn26, line 1: Shouldn't the existing THRS components also be moved here?

zero-ui/src/router/thrs/index.ts
```diff
@@ -20,6 +20,15 @@ const thrsChildRoutes: RouteRecordRaw[] = [
     },
     component: () => import("@/views/thrs/SimulationControls.vue"),
   },
+  {
+    path: "overview",
```
Boudewijn26, line 24: How about just `""`, feels more resty?

Boudewijn26: Ah, ok

PR #46, opened 2025-09-25
Update thrusters module an implement new thrusters control 

aljevandam:
New thrusters control needs proper testing. This PR also fixes some bugs in the controllers.


zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -49,7 +49,7 @@
 
 CONTROL_PARAMS = {"THRUSTERS": ThrustersParameters()}
 
-CONTROLS = {"THRUSTERS": ThrustersControl(CONTROL_PARAMS["THRUSTERS"])}
+CONTROLS = {"THRUSTERS": ThrustersControl(CONTROL_PARAMS["THRUSTERS"], datetime.now())} #TODO: should retreive time from somewhere else?
```
Boudewijn26, line 52:
```suggestion
CONTROLS = {"THRUSTERS": ThrustersControl(CONTROL_PARAMS["THRUSTERS"], datetime.now())} #TODO: should init control in simulation run instead of statically
```

zero-thrs-control/src/thrs/control/controllers.py
```diff
@@ -49,92 +52,132 @@ def setpoint(self, value: SetpointUnit):
         self._pid.setpoint = value
 
     def __call__(self, measurement: SetpointUnit, time: datetime) -> ValueUnit:
+        if self._reset_time: #TODO: deal with dt is simulation versus reality. In simulation, passing dt = tick_duration works. In real time, having actual time would be nice..
+            self._reset_time = False
```
Boudewijn26, line 56: This is just no-opping?

I think it's fine to forcefully pass a time to simple-pid. In real time `time` would be the start of the control cycle, which is indicative of the "now" as that is the best indication of the recentness of the sensor values. The main issue is that `enable` should also force a time (we could also have the cycler pass a time function instead of just a time, then we could just pass that along to the pid and be done with it)

zero-thrs-control/src/thrs/control/controllers.py
```diff
@@ -49,92 +52,132 @@ def setpoint(self, value: SetpointUnit):
...
         self._pid.time_fn = lambda: time.timestamp()
         pid_result = cast(ValueUnit | None, self._pid(measurement))
-        return pid_result if pid_result is not None else self._initial
+        return (
+            pid_result if pid_result is not None else self._initial
+        )  # do we want initial back or None?
```
Boudewijn26, line 61: `pid_result` is only `None` if it was never enabled, so then `initial` makes sense

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -44,91 +41,103 @@ class ThrustersSensorValues(ThrsModel):
...
     thrusters_flowcontrol_aft: Annotated[
-        sensor.Valve, component_meta(yard_tag="5001064-01", valve_type="flowcontrol")
+        sensor.Valve, component_meta(yard_tag="50001215", valve_type="flowcontrol")
     ]
     thrusters_flowcontrol_fwd: Annotated[
-        sensor.Valve, component_meta(yard_tag="5001064-02", valve_type="flowcontrol")
+        sensor.Valve, component_meta(yard_tag="50001064-02", valve_type="flowcontrol")
     ]
     thrusters_shutoff_recovery: Annotated[
-        sensor.Valve, component_meta(yard_tag="5001069-10", valve_type="shutoff")
```
Boudewijn26, line 53: Oh god, how did we not notice for so long

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -44,91 +41,103 @@ class ThrustersSensorValues(ThrsModel):
...
         sensor.PressureSensor, component_meta(yard_tag="50001097-02")
     ]
-
     thrusters_aft: Annotated[
-        sensor.Thruster, component_meta(yard_tag="15001001", included_in_fmu=False)
+        sensor.Thruster, component_meta(yard_tag="150001001", included_in_fmu=False)
     ]
     thrusters_fwd: Annotated[
-        sensor.Thruster, component_meta(yard_tag="15001002", included_in_fmu=False)
+        sensor.Thruster, component_meta(yard_tag="150001002", included_in_fmu=False)
```
Boudewijn26, line 71: This seems like this was correct previously

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -44,91 +41,103 @@ class ThrustersSensorValues(ThrsModel):
...
+        sensor.TemperatureSensor, component_meta(included_in_fmu=False)
+    ]:
+        total_flow = self.thrusters_flow_aft.flow.value + self.thrusters_flow_fwd.flow.value
+        if total_flow > 0.0:
+            averaged_return_temperature = (
+                self.thrusters_temperature_aft_return.temperature.value * self.thrusters_flow_aft.flow.value
+                + self.thrusters_temperature_fwd_return.temperature.value * self.thrusters_flow_fwd.flow.value
+            ) / total_flow
+        else:
+            averaged_return_temperature = 0.0  # TODO: 0 ok?
```
Boudewijn26, line 91: I'd having trouble picking between `None` and just `temp_aft + temp_fwd / 2`. We don't check flow anywhere to validate temperatures. But at the same time making any assumptions is tricky and having `None` to force handling that case seems sensible

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -44,91 +41,103 @@ class ThrustersSensorValues(ThrsModel):
...
+
+        return sensor.TemperatureSensor(temperature=Stamped.stamp(averaged_return_temperature))
```
Boudewijn26, line 93: I think this should have the `min` or `max` of `temp_aft, temp_fwd, flow_aft, flow_fwd`

zero-thrs-control/tests/modules/thrusters/test_control.py
```diff
@@ -200,26 +180,29 @@ async def test_heat_dump_with_hot_sea(
...
-        assert result.sensor_values.thrusters_flow_aft.flow.value == approx(
-            control.parameters.recovery_thruster_flow, abs=1
-        )
-        assert result.sensor_values.thrusters_flow_fwd.flow.value == approx(
-            control.parameters.recovery_thruster_flow, abs=1
+        assert (
+            result.sensor_values.thrusters_temperature_recovery.temperature.value
+            == approx(
+                control.parameters.recovery_temperature,
+                abs=2,  # tune control to decrease error margin and warm-up time
```
Boudewijn26, line 200:
```suggestion
abs=2, # TODO: tune control to decrease error margin and warm-up time
```

zero-thrs-control/tests/modules/thrusters/test_control.py
```diff
@@ -114,44 +107,31 @@ async def test_recovery(control: ThrustersControl, executor: SimulationExecutor)
...
-        if (
-            result.sensor_values.thrusters_temperature_fwd_return.temperature.value
-            <= control.parameters.recovery_mix_setpoint
-        ):
-            assert control_values.thrusters_mix_fwd.setpoint.value == approx(
-                Valve.MIXING_B_TO_AB, abs=1e-2
-            )
+        assert control_values.thrusters_mix_recovery.setpoint.value == approx(
+            Valve.MIXING_B_TO_AB,
+            abs=1e-1,  # TODO: why 2 necessary?
```
Boudewijn26, line 125: 2?

PR #47, opened 2025-09-25
ZERO-848: add zuiselect component

null-33:
ZERO-847: add dropdown component

Also includes styling updates and fixes for input and button.


null-33:
All of the documentation is generated and updated by CoPilot, which is why optimisation for readability is not much of a concern. As long as the rendered documentation page itself is easy to read, I'm personally okay with unoptimised source code if AI does maintenance.

PR #48, opened 2025-09-26
Minor fixes to gold layer queries

stefan-hofman:
Two minor fixes for the gold layer queries


data/risingwave/models/30_gold/energy/electrical_hourly.sql
```diff
@@ -4,8 +4,8 @@ SELECT
     hour,
     energy_wh
 FROM ( (
-    -- Calculate energy (Wh) of completed hours (not running hour) proportionally by calculating average values per minute to determine which minutes have data.
-    -- The energy is then proportional to the amount of minutes with data related to the total amount of minutes in the hour.
+    -- Calculate energy (Wh) of completed hours (not current hour) proportionally by calculating average values per minute to determine which minutes have data.
+    -- The energy value is then extrapolated to the total amount of minutes in the hour.
```
Boudewijn26, line 8: I don't think the logic is extrapolating or am I missing something? This is just averaging power per minute into energy per minute and then summing that, right?

Boudewijn26 (commented):
See https://github.com/foundation-zero/zero/pull/48#discussion_r2387189206

data/risingwave/models/30_gold/energy/electrical_hourly.sql
stefan-hofman: You are absolutely right! My bad.
Indeed the passed hour values are simply summed up and divided by 60.

Boudewijn26 (approved):

PR #49, opened 2025-09-26
Hook API to simulation through MQTT

Boudewijn26:


zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -356,6 +359,44 @@ class SetValuesMessageWithModel(
             except ResetError:
                 pass  # Reset the sequencer and start over
 
+    async def run_blind(self, mode: Modes):
```
dvweersel, line 362: Missing docstring. Unclear to me what the difference is between normal run and `run_blind`.

The arguments (args.type) in the cli also don't really paint a clear picture.
```
        if args.blind:
            await controls.run_blind(args.type)
```

zero-thrs-control/src/thrs/graphql/messaging.py
```diff
@@ -0,0 +1,64 @@
+from asyncio import Future
+from typing import Coroutine
+from aiomqtt import Client as MqttClient, Message
+
+from thrs.input_output.base import ThrsModel
+from thrs.input_output.modules.thrusters import ThrustersControlValues
+
+
+class Messaging[SensorValues: ThrsModel, ControlValues: ThrsModel]:
+    def __init__(
```
dvweersel, line 10: I'd prefer a docstring for classes explaining their behaviour.

zero-thrs-control/README.md
```diff
@@ -28,7 +28,7 @@ poetry install
 Run the API with
 
 ```bash
-poetry run strawberry server thrs.graphql.strawberry
+poetry run fastapi dev src/thrs/graphql/strawberry.py
```
dvweersel, line 31: Why not add a [project.scripts] and run it like that? 

I think any command involving `src` is probably incorrect and should either be replaced by a script entrypoint or run from the packaged installation created by Poetry."

zero-thrs-control/src/thrs/control/manual.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Callable
+from thrs.classes.control import Control, ControlResult
+from thrs.input_output.base import ThrsModel
+
+
+class EmptyParameters(ThrsModel):
+    pass
+
+
+class ManualControl[SensorValues: ThrsModel, ControlValues: ThrsModel](
```
dvweersel, line 11: Docstring for reasons mentioned earlier

zero-thrs-control/README.md
Boudewijn26: Because `[project.scripts]` doesn't support shell commands like package.json does and I find adding a python script to run the shell command a bit too much

PR #60, opened 2025-10-02
Add schema for simulation control, inputs and parameters

Boudewijn26:


zero-thrs-control/tests/modules/pcm/test_control.py
```diff
@@ -34,9 +34,7 @@ async def test_idle(control: PcmControl, executor: PcmExecutor):
     )
     assert control.mode == "idle"
     assert pcm_flow == approx(0.0, abs=0.1)
-    assert result.simulation_inputs.get_values_at_time(
-        executor.time()
-    ).pcm_producers_supply.flow.value == approx(
+    assert result.simulation_inputs.pcm_producers_supply.flow.value == approx(
```
Boudewijn26, line 37: `result.simulation_inputs` is already `get_values_at_time`. With my refactor `get_values_at_time` no longer returns something on which you can call `get_values_at_time` (which I think is valid)

aljevandam (approved):

PR #62, opened 2025-10-02
Validate yard tags

Boudewijn26:
ZERO-844


zero-thrs-control/src/thrs/input_output/modules/pvt.py
```diff
@@ -19,9 +19,6 @@ class PvtSensorValues(ThrsModel):
     pvt_temperature_main_aft_supply: Annotated[
         sensor.TemperatureSensor, component_meta(yard_tag="50001038-22")
     ]
-    pvt_temperature_main_aft_return: Annotated[
-        sensor.TemperatureSensor, component_meta(yard_tag="50001038-04")
-    ]
```
Boudewijn26, line 24: Duplicated, probably during a rebase

PR #65, opened 2025-10-03
Collect computed sensor values

Boudewijn26:
ZERO-859


zero-thrs-control/src/thrs/input_output/fmu_mapping.py
```diff
@@ -16,43 +14,13 @@ def groupby(iterable, key):
...
-            if included_in_fmu(field)
-        }
-
-    vals = [
-        _values_for_component(component_name, getattr(model, component_name))
-        for component_name, field in model.model_fields.items()
-        if included_in_fmu(field)
-    ]
-    return reduce(operator.ior, vals, {})
+    ) # type: ignore
```
Boudewijn26, line 23: I don't think this `type: ignore` is doing anything

zero-thrs-control/src/thrs/simulation/io_mapping.py
```diff
@@ -29,8 +63,8 @@ def tick(
         tick_duration: timedelta,
     ) -> tuple[S, O, dict[str, Any]]:
         fmu_inputs = {
-            **build_inputs_for_fmu(control_values),
-            **build_inputs_for_fmu(simulation_inputs),
+            **flatten_model_values(control_values, True),
```
Boudewijn26, line 66: I very much prefer explicitly naming boolean flags (much more readable) or using string literals

Boudewijn26 (commented):
Generally agree this approach is better. Just some styling pointers

aljevandam (approved):

PR #280, opened 2026-04-08
Add simulation inputs and fix generic type

aljevandam:


null-33:
We should consider some refactoring in the future to reduce the amount of boilerplate creeping in. That's not for this PR though.

null-33 (approved):

PR #284, opened 2026-04-12
Add helm chart to prop test

Boudewijn26:


zero-prop-test/charts/values.yaml
```diff
@@ -0,0 +1,37 @@
+# Required: IP address this service is reachable at from TwinCAT.
+# Used as the external IP of the LoadBalancer service.
+# TWINCAT_SELF_NETID is derived by appending ".1" (e.g. 10.0.0.5 -> 10.0.0.5.1).
+selfIp: ""
```
florissmit10: Why not just name this externalIP(s) like with all our other apps? Also I would put this under twincat: block since that is what it exposes.

zero-prop-test/src/zero_prop_test/settings.py
```diff
@@ -1,23 +1,47 @@
...
     twincat_username: str
     twincat_password: str
     twincat_route_name: str
-    mqtt_host: str
-    mqtt_port: int
-    mqtt_username: str
-    mqtt_password: str
+
+
+class Settings(MqttSettings, IoLinkSettings, ModbusSettings, TwinCatSettings):
```
florissmit10, line 42: What is the purpose of splitting out these settings?

Boudewijn26: Being able to require only the settings that are required for the command (so you can run the twincat data collection without having to supply the io link and modbus settings)

florissmit10 (approved):

PR #285, opened 2026-04-12
Add prop test to vector

Boudewijn26:


vector/config.yaml
```diff
@@ -10,7 +10,9 @@ sources:
     port: ${MQTT_PORT}
     decoding:
       codec: json
-    topic: "marpower/#"
+    topic: 
+      - "marpower/#"
+      - "prop-test/#"
```
florissmit10, line 15: Dont we need seperate input blocks for each kind of topic? This sends all prop-test/ data also through the Marpower transformations right?

Boudewijn26: Good point, but some of the processing (flattening and tabling) also applies to the prop test

I think running everything through the same transformation keeps things simpler than having separate pipelines, so I renamed it to `processing`

florissmit10: Simpler? Or more complex because you have to test that stuff is not proccessed when it should not be.
Fine for now, but I expect that it will be easier to have seperate pipelines in the future, where it is possible to reuse the functions if they are generic.

docker-compose.yml
```diff
@@ -213,6 +213,50 @@ services:
...
+  # ---- data collection ----
+  vector:
+    image: timberio/vector:0.54.0-debian
+    profiles:
+      - data-collection
+    command: -c /etc/vector-config.yaml
+    environment:
+      MQTT_HOST: vernemq
+      MQTT_PORT: 1883
+      MQTT_USER: ${PG_USER}
```
florissmit10: PG_USER?

florissmit10 (approved):

florissmit10:
Approved with comment

PR #288, opened 2026-04-13
Use fixed grid for dynamics, rows for sails, remove superfluous conta…

aljevandam:
…iners from docker compose

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -332,29 +332,6 @@ services:
     environment:
       <<: *app_env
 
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
       timeout: 5s
       retries: 5
 
-  postgres:
+  postgres:
```
Boudewijn26: No need for trailing whitespace

Boudewijn26 (approved):

PR #289, opened 2026-04-13
Feature/MQTT update

stefan-hofman:
Updated MQTT topic structure for some systems and updated IO lists.

[ZERO-1603](https://foundationzero-team.atlassian.net/browse/ZERO-1603)

[ZERO-1603]: https://foundationzero-team.atlassian.net/browse/ZERO-1603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


florissmit10:
When do we expect Marpower to incorporate these changes?
I see a risk to change the RW models without Marpower's implementation on board zero. If we merge this we need to wait for marpower's update to apply the models

stefan-hofman:
> When do we expect Marpower to incorporate these changes? I see a risk to change the RW models without Marpower's implementation on board zero. If we merge this we need to wait for marpower's update to apply the models

Isn't this always an issue for new changes when we rely on Marpower? I have sent the IO list fixes to Marpower and they confirmed all changes so I expect they will implement them soon.

florissmit10 (approved):

PR #295, opened 2026-04-21
Feature/Move Fresh water dashboard to greptime

stefan-hofman:
- Added greptime databases to grafana docker configuration
- Updated Fresh water dashboard to query Greptime DB instead of RisingWave.

[ZERO-1669](https://foundationzero-team.atlassian.net/browse/ZERO-1669)

[ZERO-1669]: https://foundationzero-team.atlassian.net/browse/ZERO-1669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


grafana/greptime.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: 1
+
+datasources:
+  - name: greptimedb-zero
+    type: info8fcc-greptimedb-datasource
+    access: proxy
+    jsonData:
+      host: http://100.74.188.47:4000
```
Boudewijn26: I'd prefer using the `greptimedb-zero.tail0b4840.ts.net` hostname over the IP (hostname is more readable)

Boudewijn26 (approved):

zero-dashboards/Dockerfile
```diff
@@ -36,3 +36,5 @@ RUN if [ ! -z "${GF_INSTALL_PLUGINS}" ]; then \
 
 COPY ./dashboard-provider.yaml /etc/grafana/provisioning/dashboards/dashboard-embedded-provider.yaml
 COPY ./dashboards /embedded-dashboards
+
+COPY ./greptime.yaml /etc/grafana/provisioning/datasources/greptime.yaml
```
florissmit10: This is for local development using docker-compose right? I would prefer mounting this in the container in the docker-compose.yaml over this method. This container will be used on zero as wel, and there the data sources will be different

grafana/greptime.yaml
florissmit10: Agree, this ip might change

stefan-hofman: Do you want this for sub-zero and singel as well? Or only for zero?

Boudewijn26: All of them

zero-dashboards/Dockerfile
stefan-hofman: Moved yaml to `grafana` folder and mounting is done in docker-compose.yml

grafana/greptime.yaml
stefan-hofman: Because of problems with the DNS names this change will be done later

Boudewijn26: We are experiencing some issues with the zeros, so that might explain any problems you might be having

florissmit10 (approved):

grafana/greptime.yaml
Boudewijn26: Those issues have now been resolved

florissmit10: I tested this out locally and for me the dns name does work.

PR #296, opened 2026-04-22
ZERO-1536: refactor domestic to fit top nav

null-33:


zero-ui/src/modules/common/components/navigation/NavTabs.vue
```diff
@@ -0,0 +1,37 @@
...
+import { useRoute } from "vue-router";
+import { NavItem } from "../../types";
+import RouterLinkWithFallback from "../RouterLinkWithFallback.vue";
+
+const { name } = toRefs(useRoute());
+defineProps<{ items: NavItem[] }>();
+
+const { t } = useI18n();
+
+const getRootRoute = (route: string = "") => route.split(":")[0];
```
Boudewijn26: Can we move this logic somewhere more routery?

Boudewijn26 (approved):

PR #297, opened 2026-04-23
ZERO-1693: implement FlowValve and SwitchValve and document AI instructions

null-33:


zero-ui/docs/mimics/authoring-workflow.md
```diff
@@ -0,0 +1,116 @@
+# Mimic Component Authoring Workflow
+
+This document is the canonical process for building new mimic components from Figma into Zero UI.
```
Boudewijn26:
```suggestion
This document how to build mimic components from Figma into Zero UI.
```

Boudewijn26: Is there a workflow for updating the component based on changes in Figma?

zero-ui/docs/mimics/authoring-workflow.md
```diff
@@ -0,0 +1,116 @@
...
+
+Use this workflow for every new mimic component to keep implementation quality and consistency high.
```
Boudewijn26: Less is more? If it's documented, it's the way it should be used. No need to state what should be obvious

zero-ui/docs/mimics/authoring-workflow.md
Boudewijn26, line 1: This whole document is rather verbose and wordy (canonical, non-negotiable etc). I think that gets in the way of the actual message. Could you condense it?

zero-ui/docs/mimics/authoring-workflow.md
```diff
@@ -0,0 +1,116 @@
...
+These components should be small, composable primitives intended for larger mimic scenes.
+
+## Non-Negotiable Rules
+
+- **Pure SVG template only** for all components.
+- **No HTML wrappers** (`div`, `span`, etc.) around components.
+- **One base geometry** when states are rotational variants.
+- **Semantic tokens only** for color (no hardcoded hex/rgb in final component).
+- **Typed state prop** with explicit union types.
+- **Exact Figma geometry** (exported vectors), avoid hand-drawn approximations.
```
Boudewijn26:
```suggestion
- **Exact Figma geometry** (exported vectors), do not use hand-drawn approximations.
```

zero-ui/docs/mimics/authoring-workflow.md
```diff
@@ -0,0 +1,116 @@
...
+- **Documentation required** for every component with examples and token notes.
+- **Zero diagnostics** on all touched files before merge.
+- **All text labels as SVG text** unless exact glyph shapes are required, then use path data from Figma.
+- **Created files must be formatted and linted** according to project standards.
+- **State to colour mapping should be done by a computed property using a Record<Type, string>** for maintainability and clarity.
```
Boudewijn26: I thought we were en-US (colour)

zero-ui/docs/mimics/flow-valve.md
Boudewijn26, line 1: I would prefer calling it a FlowcontrolValve to keep it consistent with the API

zero-ui/src/modules/thrapp/mimics/components/flow-valve/FlowValve.vue
Boudewijn26, line 1: Does it make sense to have completely separate components for each valve type or would it make more sense to have a single `Valve` component with different types? I'd lean towards the latter myself (in line with component libraries like Vuetify which have a button with flat, raised, etc)

zero-ui/docs/mimics/authoring-workflow.md
null-33: I will add an introduction to this document to clarify that this is an instruction document for {insert favourite AI}. It might be rather verbose but it is not really meant for us humans. I asked CoPilot to write this for himself so that we always get more or less the same sort of output when asked to extract a component from Figma.

zero-ui/src/modules/thrapp/mimics/components/flow-valve/FlowValve.vue
null-33: It might not be, although states and variants are also dependant on the type (if we include the three-way valve), which would create another kind of complexity. I will see if I can do this for the SwitchValve and FlowControlValve to begin with.

zero-ui/docs/mimics/authoring-workflow.md
null-33: CoPilot is, but this was my own addition and I sometimes tend towards en-GB 😅.

zero-ui/docs/mimics/authoring-workflow.md
null-33: Not right now, but I will add this when the opportunity presents itself.

Boudewijn26: I was under the impression that that was part of the challenge to solve: doing a one-time import is simple enough. Handling updates is where things get tricky

zero-ui/docs/mimics/authoring-workflow.md
null-33: I'm leaning towards keeping it in, since CoPilot added this for itself. I find that treating it like a child and keep repeating yourself yields better results.

zero-ui/docs/mimics/authoring-workflow.md
null-33: I've added a separate instructions file for maintenance and tested this with adding the three way valve to the Valve component. While this still required some refactoring, the overall process was successful. I expect simple design changes to be fully automated.

zero-ui/src/modules/thrapp/mimics/components/flow-valve/FlowValve.vue
null-33: I've merged the two components and added the three way valve as well.

Boudewijn26 (approved):

PR #302, opened 2026-04-24
Add charts to zero-hull-temperature

Boudewijn26:


zero-hull-temperature/charts/values.yaml
```diff
@@ -0,0 +1,40 @@
+replicaCount: 1
+
+image:
+  repository: europe-west1-docker.pkg.dev/common-449414/common/zero-hull-temperature
+  pullPolicy: IfNotPresent
```
florissmit10, line 5: Better to do always. Then it checks if the sha has changed every time. If it is still the same it will not download it again.

zero-hull-temperature/charts/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+  port: 1883
+  username: hull-temperature
+  existingSecret: ""
+  existingSecretPasswordKey: password
+
+modbus:
+  host: ""
+  port: 502
+
+run:
```
florissmit10: It is cosmetic, but I dont really like the use of run here and in the ENV variables. I prefer nothing, or the application name.

florissmit10 (approved):

zero-hull-temperature/charts/values.yaml
Boudewijn26: We're using `IfNotPresent` everywhere. Depending instead on the version update of the release to do the lookup. I'd be curious what would happen if a pod needs to be created when the boat has no internet and the `pullPolicy` is always

PR #303, opened 2026-04-24
Use hasura to do migrations

Boudewijn26:


data/README.MD
```diff
@@ -20,9 +20,9 @@ docker compose --profile zero up
 This will set up the following relevant services:
  - VerneMQ: MQTT client
  - RisingWave: Streaming database
+ - Postgres: managed through Hasura migrations
```
aljevandam, line 23: nitpicking - but this is not consistent with the above. Given that for other services this decribes what they are (not how they are managed), I'd expect something like "Database [that holds x/for metadata/for static data], managed through Hasura migrations"

hasura/migrations/zero/20260424000300_loads_cases/down.sql
```diff
@@ -0,0 +1,9 @@
+TRUNCATE TABLE
```
aljevandam, line 1: why use truncate here and drop in domestic control?

aljevandam (commented):
What's the motivation to use Hasura for this?

Boudewijn26:
The main motivation is that we have Hasura already, so we don't need to introduce and monitor a new tool to do this for us

hasura/migrations/zero/20260424000300_loads_cases/down.sql
Boudewijn26: [Because the `up` migration does insertions, it doesn't create any tables](https://github.com/foundation-zero/zero/blob/5e4155e20316b6ea840cd520b2075abe59156ed0/hasura/migrations/zero/20260424000300_loads_cases/up.sql)

aljevandam (approved):

PR #316, opened 2026-05-05
Add boilers alarms

aljevandam:


zero-thrs-control/src/thrs/control/modules/boilers.py
```diff
@@ -750,4 +750,173 @@ def _disable_pump_flow_control(self, sensor_values: BoilersSensorValues):
...
+        return None
+
+    @staticmethod
+    def _check_tank_level(
+        tank_number: int,
+        level: float,
+        maximum: float,
+    ) -> str | None:
+        if level > maximum:
+            return f"Tank {tank_number} level {level:.1f}L above {maximum}L"
```
Boudewijn26: I think the crew would prefer the yard tag over the tank number

Boudewijn26 (approved):

PR #319, opened 2026-05-06
Add grafana to splash screen

Boudewijn26:


zero-ui/src/modules/common/views/Grafana.vue
```diff
@@ -0,0 +1,41 @@
...
+          Set VITE_GRAFANA_URL in your environment to enable dashboard embedding.
+        </p>
+        <RouterLink
+          class="mt-5 inline-block"
+          :to="{ name: 'splash' }"
+        >
+          <Button variant="secondary">Back to splash</Button>
+        </RouterLink>
+      </div>
+    </section>
```
Boudewijn26, line 43: Not sure if we want to keep this, can be handy for development if something goes awry, but shouldn't happen normally

zero-ui/src/modules/common/views/Grafana.vue
null-33, line 1: Maybe add these strings to i18n in case it's more permanent.

null-33 (approved):

PR #323, opened 2026-05-11
Add boat executor

Boudewijn26:


zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -22,27 +20,21 @@
 logger = logging.getLogger(__name__)
 
 
-class MqttExecutor[O: SimulationValues](Executor[CombinedValues, CombinedValues]):
-    # Controller client listens to sensor values and publishes control values (our control logic)
-    # Environment client listens to control values and publishes sensor values (mimicking the PLC)
+class MqttControlExecutor(Executor[CombinedValues, CombinedValues]):
```
aljevandam: As this does not have a inner, and takes the control values as input, this is basically just an MQTT connector (right?) 

I'd suggest to call it that (future PR)

zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -106,6 +62,75 @@ async def _publish_by_mapping[T](
...
+        return self._start_time
+
+    def time(self) -> datetime:
+        return datetime.now()
+
+
+class MqttSimulationExecutor(Executor[CombinedValues, CombinedValues]):
+    def __init__(
+        self,
+        inner: Executor,
```
aljevandam, line 112: Strictly speaking the inner could be required to be a SimulationExecutor?

zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -106,6 +62,75 @@ async def _publish_by_mapping[T](
...
+        environment_client: Client,
```
aljevandam: and this could be simulation_client or just client?

zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -22,27 +20,21 @@
...
 
 
-class MqttExecutor[O: SimulationValues](Executor[CombinedValues, CombinedValues]):
-    # Controller client listens to sensor values and publishes control values (our control logic)
-    # Environment client listens to control values and publishes sensor values (mimicking the PLC)
+class MqttControlExecutor(Executor[CombinedValues, CombinedValues]):
     def __init__(
         self,
-        inner: Executor,
         controller_client: Client,
```
aljevandam: this can just be client?

zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -106,6 +62,75 @@ async def _publish_by_mapping[T](
...
+
+    @property
+    def start_time(self) -> datetime:
+        return self._start_time
+
+    def time(self) -> datetime:
+        return datetime.now()
+
+
+class MqttSimulationExecutor(Executor[CombinedValues, CombinedValues]):
```
aljevandam: In future PR, I'd suggest we have an MqttConnector and a SimulationExecutor (or just Simualtion) separately

zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -106,6 +62,75 @@ async def _publish_by_mapping[T](
...
+        )
+
+    async def run(self):
+        self._running = True
+        try:
+            await self._listen_to_sensors()
+        finally:
+            self._running = False
+
+    async def tick(
```
aljevandam, line 86: isn't it strange that the control_values are expected as an input here. I'd say this thing should listen to sensor values on MQTT and then call the control to get control_values an publish those on MQTT.

Currently, this is just publishing control values to MQTT

zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -189,6 +187,54 @@ def time(self) -> datetime:
...
+        )
+
+    async def start(self):
+        await self._control_executor.start()
+        await self._simulation_executor.start()
+
+    async def run(self):
+        await self._control_executor.run()
+
+    async def tick(
```
aljevandam, line 221: I would say to really have lock-step, we would have something like 

def tick()
     sensor_values = self.simulation_executor.tick(self.control_values)
     control_values = self.control.tick(self.sensor_values) 

      self.publish_control_values()
      self.publish_sensor_values() 

i.e. we'd just publish what's going on, and not have the simulation and control communicate over MQTT. 

But this might not be needed for anything...

zero-thrs-control/src/thrs/orchestration/executor.py
```diff
@@ -189,6 +187,54 @@ def time(self) -> datetime:
...
+
+    @property
+    def start_time(self) -> datetime:
+        return self._simulation_executor.start_time
+
+    def time(self) -> datetime:
+        return self._simulation_executor.time()
+
+
+BoatExecutor = MqttControlExecutor
```
aljevandam: Let's get rid of BoatExecutor

aljevandam (approved):

zero-thrs-control/src/thrs/orchestration/executor.py
Boudewijn26: Just because that's how we use it now doesn't mean that throwing in a different inner is invalid. You could nest an MqttExecutor in an MqttExecutor (as long as the topic prefices differ)

zero-thrs-control/src/thrs/orchestration/executor.py
Boudewijn26: I would disagree: this gives you a nice separation of function and semantics. You can replace `BoatExecutor` by something else and thereby adjust anything which should execute like the boat without breaking stuff that wants to send controls over MQTT

zero-thrs-control/src/thrs/orchestration/executor.py
aljevandam: what's the point of that though other than causing confusion

Boudewijn26: It's just an example, we don't know yet what other executors we might want (maybe a playback rate executor (could actually be a good idea) which again wraps another executor). Perhaps the better argument against making it a SimulationExecutor would be that it breaks abstraction. MqttSimulationExecutor has no need to know it has a SimulationExecutor as the inner. Even calling it MqttSimulationExecutor is maybe a bad idea

If you wanted a simpler interface, then you can have a class or function which creates both executors for you

zero-thrs-control/src/thrs/orchestration/executor.py
hmvp: Done

zero-thrs-control/src/thrs/orchestration/executor.py
hmvp: I disagree: There is a MqttControlExecutor (that should have been named connector of something since its job is to put the results of a module on mqtt). That it is the one we want to be running on the boat is incidental and that we use it on the boat should be named/documented in the bit of code that glues everything together for deployment.

PR #331, opened 2026-05-21
Add Graphql endpoints for controls

aljevandam:


hmvp (approved):
Still a bit to fresh to be confident enough about my review capabilties for this, but as far as I can see this looks good to me. (I assume the added controls match the io list)

null-33 (approved):
No breaking changes for the frontend.

PR #332, opened 2026-05-28
Add groups for propulsion and power storage topics

florissmit10:
Needs to be cleaned up at the IO list level, but for now this should fix the data ingest

ZERO-1820


vector/processing/0_consts.vrl
```diff
@@ -1,6 +1,7 @@
 # This file is auto-generated by zero-data
 grouped_topics = [
   "marpower/150000-propulsion/",
+  "marpower/150000propulsion/",
```
Boudewijn26: Under 150000propulsion we have the aradexes, the helm stations and the pcs itself, those shouldn't be grouped together

Boudewijn26 (changes requested):

vector/processing/0_consts.vrl
florissmit10: Ah ok, it looked like a nested topic. Will remove it from here then

Boudewijn26 (approved):

PR #344, opened 2026-06-05
Add justfile to thrs project

hmvp:


florissmit10:
I like this change, as it improves understandability of the codebase, and reduces differences between CI and local! I have seen just used in other places and I think we can use it in other projects as wel

zero-thrs-control/.vscode/settings.json
```diff
@@ -3,6 +3,7 @@
         "tests"
     ],
     "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
-}
-
+    "python.testing.pytestEnabled": true,
+    "python-envs.defaultEnvManager": "ms-python.python:poetry",
```
florissmit10, line 7: This looks windows specific, is it?

hmvp: No the `ms-` comes from the the vscode python plugin that is made by Microsoft, there is nothing windows specific in this file. (I'm on linux)

florissmit10 (approved):
LGTM, But maybe add a documentation line with a link to install instructions

PR #347, opened 2026-06-05
ZERO-1792: support new states for some components

null-33:
AKA "Pinball Mode"
Scherm­afbeelding 2026-06-05 om 15 06 09

This PR adds the orange/red outline and labels for components that are in Manual or Alarm state. This also changes the implementation of the ActuatedValve component to support all 4 variations (Switch Valve, Flow Control Valve, Mix Valve and 3-Way Switch Valve).


hmvp:
LGTM for the most part, if I quickly toggle demo mode (which enables alarms on some) Then there is some flicker because the border widths change slightly on the bigger boxes which also moves their contents slightly...

null-33:
> LGTM for the most part, if I quickly toggle demo mode (which enables alarms on some) Then there is some flicker because the border widths change slightly on the bigger boxes which also moves their contents slightly...

Good catch! I added a second container, just for the background and border, so that the content doesn't get pushed inwards when the stroke width increases.

hmvp (approved):

PR #349, opened 2026-06-05
Refactor control module definitions to make code more readable

hmvp:
Move ModuleDescriptions to the respective control modules to reduce the number of imports.
And  to make it clear what its needed for control module to exist...


Boudewijn26 (approved):
LGTM, though I would like to point out that controls aren't strictly tied to modules. You can imagine a thrusters module with a `NoopControl` or it's `ThrustersControl` inside of a `SwitchingControl`. So the constants you declared might be the canonical modules, but can (and sometimes should) be deviated from

PR #356, opened 2026-06-10
Refactor THRS orchestration

hmvp:
- Merge and rename Simulation + Cycler into Runner
-  Split Executor into Connector (handle mqtt comms) and Simulation (run FMU)
-  Add fix for extraction of fmu into tmp folder


zero-thrs-control/src/thrs/orchestration/connector.py
Boudewijn26, line 1: It might be worthwhile to have an `IdConnector` or `MemoryConnector`. That way you don't need special code to handle the MQTT-less case (or keep it in your back pocket for when you actually need it)

Boudewijn26 (approved):

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -825,20 +826,20 @@ async def _run_simulation(
                 SimulationStatusMessage(
                     mode=mode,
                     status="available",
-                    simulation_time=executor.time(),
+                    simulation_time=connector.time(),
```
aljevandam, line 829: I don't think the time should come from the connector? shouldn't it come from the simulation (since you can run the simulation for example at a x3 rate) 

But the how does the control know what time it is ?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -20,7 +19,23 @@
...
 
 
-class MqttControlConnector(Executor[CombinedValues, CombinedValues]):
+@dataclass
+class ExecutionResult[S]:
+    timestamp: datetime
+    sensor_values: S
+
+
+class Connector[S, C](Protocol):
```
aljevandam, line 28: isn't this still implying other funcitonality than purely connecting? start, tick , etc. 

I would have thought the Connector just translates to MQTT (I know this is just a rename, buit still)

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -106,10 +121,10 @@ def time(self) -> datetime:
         return datetime.now()
 
 
-class MqttSimulationConnector(Executor[CombinedValues, CombinedValues]):
+class MqttSimulationConnector(Connector[CombinedValues, CombinedValues]):
     def __init__(
         self,
-        inner: Executor,
+        inner: "Simulation",
```
aljevandam, line 127: this can now be called simulation instead of inner?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -187,53 +202,54 @@ def time(self) -> datetime:
...
     async def run(self):
-        await self._control_executor.run()
+        await self._control_connector.run()
 
     async def tick(
         self, control_values: CombinedValues
     ) -> ExecutionResult[CombinedValues]:
-        await self._control_executor.tick(control_values)
-        return await self._simulation_executor.tick(control_values)
+        await self._control_connector.tick(control_values)
```
aljevandam, line 239: control should get sensor values? still weir in control connector

aljevandam (approved):

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: Yes

zero-thrs-control/src/thrs/cli/simulation_controls.py
hmvp: Yeah this is a weird artefact that should be resolved

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: start is because the subscription is handled continously and tick should be renamed

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: yes

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: This is only the send out part and does not do the control logic

PR #358, opened 2026-06-10
Some small cleanups in thrs

hmvp:


aljevandam (approved):

zero-thrs-control/src/thrs/control/modules/high_temperature.py
Boudewijn26, line 1: Can you motivate this deletion? I thought it'd be nice to have a clear distinct place where the high temperature module lives

hmvp: It makes refactoring easier, I suspect that it might come back at some point, but now its a unneeded exception

PR #362, opened 2026-06-12
Move Mqtt related stuff to connector.py and Simulation related stuff to simulation.py

hmvp:
To be able to move the mqtt related stuff out of the other logic I slightly changed where the mappings are generated and change things so we now communicate the sensor and control values classes instead of the mappings

### DIFF
--- a/zero-thrs-control/src/thrs/cli/simulation_controls.py
+++ b/zero-thrs-control/src/thrs/cli/simulation_controls.py
@@ -758,7 +758,9 @@ class SimulationControls:
                 self._control_client,
                 self._sensor_client,
                 self._topic_prefix,
-                modules,
+                modules.sensor_values_mqtt_mapping,
+                modules.control_values_mqtt_mapping,
+                modules.simulation_output_mqtt_mapping,
             )
             runner = Runner(connector, control, modules.alarms())  # type: ignore
 
--- a/zero-thrs-control/src/thrs/cli/simulation_controls.py
+++ b/zero-thrs-control/src/thrs/cli/simulation_controls.py
@@ -194,7 +194,7 @@ type Modes = Literal[
     "thrusters", "pvt", "pcm", "consumers", "high_temperature", "boilers"
 ]
 
-MODES: dict[Modes, tuple[str, CombinedModule]] = {
+MODES: dict[Modes, tuple[str, CombinedModule, str]] = {
     "thrusters": (
         thrusters_path,
         CombinedModule(
@@ -203,8 +203,8 @@ MODES: dict[Modes, tuple[str, CombinedModule]] = {
             },
             ThrustersSimulationInputs,
             ThrustersSimulationOutputs,
-            control_topic_suffix=settings.mqtt_control_topic_suffix,
         ),
+        settings.mqtt_control_topic_suffix,
     ),
     "pvt": (
         pvt_path,
@@ -212,8 +212,8 @@ MODES: dict[Modes, tuple[str, CombinedModule]] = {
             {"pvt": PVT_MODULE_DESCRIPTION},
             PvtSimulationInputs,
             PvtSimulationOutputs,
-            control_topic_suffix=settings.mqtt_control_topic_suffix,
         ),
+        settings.mqtt_control_topic_suffix,
     ),
     "pcm": (
         pcm_path,
@@ -221,8 +221,8 @@ MODES: dict[Modes, tuple[str, CombinedModule]] = {
             {"pcm": PCM_MODULE_DESCRIPTION},
             PcmSimulationInputs,
             PcmSimulationOutputs,
-            control_topic_suffix=settings.mqtt_control_topic_suffix,
         ),
+        settings.mqtt_control_topic_suffix,
     ),
     "consumers": (
         consumers_path,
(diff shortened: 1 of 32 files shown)


zero-thrs-control/src/thrs/input_output/model_builder.py
```diff
@@ -49,9 +49,11 @@ async def wait_for_result(self) -> T:
 
 
 class CombinedModelBuilder(ModelBuilder[CombinedValues]):
-    def __init__(self, clss: dict[str, type[ThrsValues]]):
+    def __init__(self, clss: type[ThrsValues]):
         self._model_builders: dict[str, ModelBuilder[ThrsValues]] = {
-            name: PartialModelBuilder(cls) for name, cls in clss.items()
+            name: PartialModelBuilder(field.annotation)
+            for name, field in clss.model_fields.items()
+            if field.annotation and issubclass(field.annotation, ThrsValues)
```
Boudewijn26: I'd consider moving this logic into a function somewhere. Pydantic field reflection has the tendency to evolve over time and the last thing you'll want is to have it duplicated across the codebase

zero-thrs-control/src/thrs/orchestration/module.py
```diff
@@ -238,25 +131,24 @@ def __init__(
...
+            **{module: desc.sensor_values_cls for module, desc in modules.items()},  # type:ignore
+            __base__=ThrsValues,
         )
-        self._control_mqtt_mapping = ModuleMqttMapping(
-            {module: desc.control_values_cls for module, desc in modules.items()},
-            topic_suffix=control_topic_suffix,
+        self._control_values_cls = create_model(
+            "CombinedControlValues",
+            **{module: desc.control_values_cls for module, desc in modules.items()},  # type:ignore
+            __base__=ThrsValues,
```
Boudewijn26: Do you expect to want to use some of Pydantic's features on these classes? I don't immediately see why they need to be Pydantic models

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -194,7 +194,7 @@
     "thrusters", "pvt", "pcm", "consumers", "high_temperature", "boilers"
 ]
 
-MODES: dict[Modes, tuple[str, CombinedModule]] = {
+MODES: dict[Modes, tuple[str, CombinedModule, str]] = {
```
Boudewijn26, line 197: Not necessarily a deal breaker for now, but this is reaching the end of what a tuple should hold before becoming a dataclass for clarity

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -98,7 +199,7 @@ async def run(self):
         finally:
             self._running = False
 
-    async def tick(
+    async def send_and_receive(
```
Boudewijn26: How about `transceive`? If [transceivers](https://en.wikipedia.org/wiki/Transceiver) are a thing, I think the verb `transceive` is valid

zero-thrs-control/src/thrs/cli/simulation_controls.py
hmvp: Yeah i would prob move the control prefix out of that MODES thing. But wanted to  touch as little as possible in that class for now

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: Yeah!

zero-thrs-control/src/thrs/orchestration/module.py
hmvp: It could have been a dict, probably. The main goal was not to "communicate" in Mqtt related classes

But in this way the interface of combined module mimics that of ModuleDefinition close enough that at some point they may be interchangeable and thus running a single control module or multiple is easy

zero-thrs-control/src/thrs/input_output/model_builder.py
hmvp: Yeah, there are a few places where we use similar logic... I am not sure if we can move those into a single function or similar. But I expect those part to change in the near future (depending on how we handle CombinedModule)

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: Changed

zero-thrs-control/src/thrs/input_output/model_builder.py
Boudewijn26: I'm not necessarily advocating to move all reflection to one place, but to have this specific case covered. But that can also be saved for when it needs to be reused

hmvp: Changed with added commit

zero-thrs-control/src/thrs/orchestration/module.py
hmvp: So having thought about this... It does not make sense if I only do this... The tought was that CombinedControl etc all lie about their types.  Control[] etc. assume ThrsValues subclasses for their generic parameters but CombinedModule breaks that promise by using CombinedValues.. It works but is not correct type-wise... This was an attempt to fix that and make CombinedModule type compatible with ModuleDescription. But if we go that route (and I am not sure that is a correct thing to do) then All of CombinedValues should disappear  and I did not do that...

Boudewijn26: I think the desire to be able to have a CombinedModule be describable with a ModuleDescription might come from the idea that higher level controls should be CombinedModules. I would argue against that, higher level controls can just be their own control with their own sensor values and control values (though it would make sense for the sensor values to be something combined). 

At the same time there is nothing wrong with broadening ModuleDescription to just have any sensor values and control values class (by dropping the constraint on ThrsValues for those). You could then consider a specialized subclass which does have that constraint

hmvp: Yeah, I changed it to a Mapping like it used to be

aljevandam (approved):

PR #367, opened 2026-06-16
Add cosimulation

aljevandam:


hmvp (approved):

zero-thrs-control/src/thrs/simulation/cosimulation.py
```diff
@@ -0,0 +1,152 @@
...
+                participant.fmu_key_input_mapping[
+                    (coupling.dest_component, coupling.dest_field)
+                ]: _src_fmu_key(coupling)
+                for coupling in participant.couplings
+            }
+            for participant in self._participants
+        ]
+
+    def __enter__(self) -> Self:
+        return self
```
Boudewijn26, line 113: I think it'd technically be more correct to also enter the `fmu`s

zero-thrs-control/src/thrs/simulation/cosimulation.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def __exit__(
+        self,
+        type_: type[BaseException] | None,
+        value: BaseException | None,
+        traceback: TracebackType | None,
+    ) -> bool:
+        for participant in self._participants:
+            participant.fmu.__exit__(type_, value, traceback)
```
Boudewijn26, line 122: [`ExitStack`](https://docs.python.org/3/library/contextlib.html#contextlib.ExitStack) might be useful to handle this (and the proper error handling)

zero-thrs-control/src/thrs/simulation/cosimulation.py
```diff
@@ -0,0 +1,152 @@
...
+                if key in participant.fmu_key_input_mapping.values()
+            }
+
+            # inject previous outputs from other participants based on the coupling
+            for dest_fmu_key, src_fmu_key in coupling.items():
+                if dest_fmu_key in participant_inputs:
+                    raise ValueError(
+                        f"Input '{dest_fmu_key}' for participant '{participant}' is being set both directly and via coupling from '{src_fmu_key}'."
+                    )
+                participant_inputs[dest_fmu_key] = self._previous_outputs[src_fmu_key]
```
Boudewijn26, line 142: Instead of setting directly in the dict, I'd suggest building 2 dicts and merging them

Then you can have a single duplicate set check through `if duplicate_keys := set(a.keys()) & set(b.keys()):`

PR #368, opened 2026-06-16
Cleanups and fixes to python notebooks

hmvp:


zero-thrs-control/src/thrs/orchestration/connector.py
aljevandam: change to simulation?

aljevandam (approved):

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: Done

PR #372, opened 2026-06-17
Some cleanup in thrs

hmvp:
- Removed unused code
- Removed time and start_time from connector since they are not used there (they are a simulator thing)
- Moved Collectors to test code


aljevandam:
Can we find a different place for simulation_runner and collector? They are not just used for testing, but also for analysis (in the notebooks).

mirdesign (approved):

PR #374, opened 2026-06-18
Send computed fields to mqtt

hmvp:
Computed fields defined on SensorValues are sent to their topic 

We do this by defining a direction for each field INPUT/OUTPUT (computed fields are automatically OUTPUT)
and then sending all output fields instead of only control values


zero-thrs-control/src/thrs/input_output/modules/dhw.py
```diff
@@ -213,15 +213,14 @@ class DhwSensorValues(ThrsValues):
     ]
 
     @computed_field(
-        json_schema_extra=component_meta(included_in_fmu=False).json_schema_extra
+        json_schema_extra=computed_meta(
+            component_type="delta_t",
+            included_in_fmu=False,
```
aljevandam: Again we run into terminology issues here I guess. This is not a controller value in the sense that it says something about the controller (eg the intergral term of a pid controller). This is more an 'enriched sensor value'. 

But perhaps controller here is meant to signal that it's been computed in the controller? 

I'd say it 'default topic' is fine, and we could prevent specifying the whole topic here. A 'publish=True' flag could also work perhaps to publish computed sensor values, and it could just be under the same topic structure as sensor values?

zero-thrs-control/src/thrs/input_output/modules/dhw.py
```diff
@@ -594,55 +677,37 @@ class DhwSimulationInputs(SimulationInputs):
     dhw_seawater_supply: simulation.TemperatureBoundary
     dhw_hotwater_demand: simulation.FlowBoundary
 
-    @computed_field(
-        json_schema_extra=component_meta(included_in_fmu=False).json_schema_extra
-    )
+    @computed_field(json_schema_extra=computed_meta(included_in_fmu=False))
     @property
     def drives_flow_recovery(self) -> sensor.FlowSensor:
```
aljevandam, line 567: The metadata here should be in the json_schema_extra and not the return type right?

zero-thrs-control/src/thrs/input_output/modules/dhw.py
aljevandam: These _could_ be controllervalues I guess, since the control computes them..but then we might move them to the control(ler) values instead of sensorvalues?

zero-thrs-control/src/thrs/input_output/base.py
```diff
@@ -21,13 +18,17 @@
 )
 
 
+class ControllerIoDirection(Enum):
```
aljevandam: Don't we have
- sensorvalues (always 'to controller')
- controlvalues (always 'from controller' and also 'to simualtion/amcs' or equivalent, with a /Command suffix
- controllervalues (or a better name) (always 'from controller' and published for the UI or for greptime)

Currently controlvalues holds control values and controller values, and we have choice where we put 'enriched sensorvalues (like a deltaT): in sensorvalues or in contollervalues? 

I just wonder whether we need to/want to to specify custom topics and IOdirecitons per field/topic-key, if we keep each field in the right place

Or in the case where we keep enriched sensor values under sensor values, a 'publish =True' could be sufficient for the control to publish them?

aljevandam: When do we have both?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -17,7 +17,19 @@
...
 
-    def split_to_topics(self, model: M) -> dict[str, str]: ...
+    def split_to_topics(
+        self, model: M, *, is_simulation: bool = False
+    ) -> dict[str, str]:
+        """Split model instance values to topics and payloads.
+
+        Args:
+            model: ThrsValues subclass with values to publish
+            is_simulation (bool, optional): Use this for simulations where you want to publish what is normally incoming. Defaults to False.
```
aljevandam: not sure if I get this - is_simulation as in Simulation or Control?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -39,24 +51,27 @@ def __init__(self, cls: type[M], topic_prefix: str, module_prefix: str):
...
         }
         self.topic_to_field = {
             self._topic(field_name, field): field_name
-            for field_name, field in cls.model_fields.items()
+            for field_name, field in cls.input_fields().items()
         }
         self._builder = PartialModelBuilder(self._cls)
 
-    def split_to_topics(self, model: M) -> dict[str, str]:
+    def split_to_topics(
```
aljevandam: I don't get this - the topics are inputs or output fields depending on whether it's a simulation? Shouldn't the MqttMapping be agnostic

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -180,20 +201,36 @@ async def _listen_to_sensors(self):
...
+        self,
+        client: Client,
+        mapping: MqttMapping[T],
+        value: T,
+        computed_values: bool = False,
     ):
         payloads = mapping.split_to_topics(value)
         for topic, payload in payloads.items():
-            topic = f"{topic}{self._control_topic_suffix_str}"
+            topic = (
```
aljevandam: yeah so if we keep eneriched sensor values in SensorValues (and don't move them to ControllerValues) then a publish=True would suffice here (and would be more clear)?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -207,15 +244,17 @@ async def run(self):
...
+        self, control_values: CombinedValues, sensor_values: CombinedValues | None
+    ) -> CombinedValues | None:
         if not self._running:
             raise Exception(
                 "MqttControlConnector not running, run() should be called in a create_task()"
             )
-        sensors = self._sensor_values_mqtt_mapping.result()
-        await self._send_control_values(control_values)
+        new_sensors_values = self._sensor_values_mqtt_mapping.result()
+        await self._send_control_values(control_values, sensor_values)
```
aljevandam: this is confusing naming right - why not have self._publish_control_values and self._publish_enriched_sensor_values separately?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -207,15 +244,17 @@ async def run(self):
...
+    async def transceive(
+        self, control_values: CombinedValues, sensor_values: CombinedValues | None
+    ) -> CombinedValues | None:
         if not self._running:
             raise Exception(
                 "MqttControlConnector not running, run() should be called in a create_task()"
             )
-        sensors = self._sensor_values_mqtt_mapping.result()
-        await self._send_control_values(control_values)
+        new_sensors_values = self._sensor_values_mqtt_mapping.result()
```
aljevandam: sensor_values (singular?)

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -207,15 +244,17 @@ async def run(self):
         finally:
             self._running = False
 
-    async def transceive(self, control_values: CombinedValues) -> CombinedValues:
+    async def transceive(
```
aljevandam: It's also still confusing to me the listening an dpublishing is put together in tranceive. For me, listen - tick - publish would be a cleaner pattern (because it handles all values from the current iteration, not listening to sensors of time t, and sending controls based on  t-1)

zero-thrs-control/src/thrs/orchestration/runner.py
```diff
@@ -92,8 +93,13 @@ def from_module(
 
     async def run(self, n_ticks: int) -> None:
         for _ in range(n_ticks):
-            sensor_values = await self._connector.transceive(self._control_values)
-            self._control_values = self._control.control(sensor_values).values
+            self._sensor_values = await self._connector.transceive(
+                self._control_values, self._sensor_values
+            )
+
+            sensor_values: S = self._sensor_values or CombinedValues(values={})  # type: ignore
```
aljevandam: would it be cleaner if the connector transieve returns a CombinedValues({}) instead of None?

zero-thrs-control/src/thrs/orchestration/runner.py
```diff
@@ -92,8 +93,13 @@ def from_module(
...
+
+            self._control_values = self._control.control(sensor_values)
```
aljevandam, line 96: I'd prefer explicietly sending the new controls here in this iteration, instead of implicietly in the next tick

zero-thrs-control/src/thrs/input_output/modules/dhw.py
hmvp: yes

hmvp:
@aljevandam Changed the PR as we discussed... I did need to add a seperate mapping for the computed values, but I think that is actually more elegant

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -635,12 +628,16 @@ def __init__(
         controls_client: MqttClient,
         control_client: MqttClient,
         sensor_client: MqttClient,
-        topic_prefix: str,
+        amcs_topic_prefix: str,
```
aljevandam: maybe amcs_topic_prefix should be sensors_topic_prefix? As we can put in either amcs or simulation topic?

zero-thrs-control/src/thrs/graphql/messaging.py
```diff
@@ -142,28 +138,28 @@ def __init__(
         parameters_cls: type[Parameters],
         mode_cls: type[Mode],
         mqtt_client: MqttClient,
+        amcs_topic_prefix: str,
```
aljevandam: these controls don't go on the amcs topic right? This is communication from UI to control?

aljevandam: 'directives'?

zero-thrs-control/src/thrs/graphql/messaging.py
```diff
@@ -285,18 +281,20 @@ def __init__(
...
             SimulationInputMessage[SimulationInputs],
-            SimulationInputMessage.subscribe_topic(),
+            f"{topic_prefix}/{SimulationInputMessage.subscribe_topic()}",
         )
         self._simulation_outputs = RawMessageReceiver(
-            SimulationValues, "simulation/outputs"
+            SimulationValues, f"{topic_prefix}/simulation/outputs"
         )
         self._simulation_inputs_cls = None
+        self._amcs_topic_prefix = topic_prefix
```
aljevandam: again I don't think amcs_topic is relevant here? This is the comm between UI and simulation?

zero-thrs-control/src/thrs/graphql/messaging.py
```diff
@@ -387,13 +385,16 @@ def __init__(
         mqtt_client: MqttClient,
         control_modules: list[ControlMessaging],
         simulation: SimulationMessaging,
+        amcs_topic_prefix: str,
```
aljevandam: same here

zero-thrs-control/src/thrs/orchestration/config.py
```diff
@@ -7,5 +7,6 @@ class Config(BaseSettings):
     )
     mqtt_host: str
     mqtt_port: int
-    mqtt_topic_prefix: str
+    mqtt_amcs_topic_prefix: str
```
aljevandam: mqqt_sensors_topic_prefix?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -17,7 +17,19 @@
...
 
-    def split_to_topics(self, model: M) -> dict[str, str]: ...
+    def split_to_topics(
+        self, model: M, *, computed_values: bool = False
+    ) -> dict[str, str]:
+        """Split model instance values to topics and payloads.
+
+        Args:
+            model: ThrsValues subclass with values to publish
+            computed_values (bool, optional): Use if you want to publish computed values
```
aljevandam: looking at the code, computed_values = True means construct the mapping ONLY for computed values right?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -145,21 +166,24 @@ class MqttControlConnector(Connector[CombinedValues, CombinedValues]):
...
         self._mqtt_client = mqtt_client
-        self._topic_prefix = topic_prefix
         self._control_topic_suffix_str = (
             f"/{control_topic_suffix}" if control_topic_suffix else ""
         )
         self._sensor_values_mqtt_mapping = ModuleMqttMapping(
-            sensor_values_clss, topic_prefix
+            sensor_values_clss, amcs_topic_prefix
         )
         self._control_values_mqtt_mapping = ModuleMqttMapping(
```
aljevandam, line 192: or mabye, looking at this, the amcs_topic_prefix should besomething very generic like thrs_topic_prefix (which can then either me marpower_blabla or simulation something

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -145,21 +166,24 @@ class MqttControlConnector(Connector[CombinedValues, CombinedValues]):
...
-            control_values_clss, topic_prefix
+            control_values_clss, amcs_topic_prefix
+        )
+        self._computed_values_mqtt_mapping = ModuleMqttMapping(
```
aljevandam, line 195: I'd expect the only_computed_values flag here in the ModuleMqttMapping init (as it goes hand in hand with the topic and class right?)

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -212,29 +254,29 @@ async def transceive(self, control_values: CombinedValues) -> CombinedValues:
...
+        return sensors_values if sensors_values else CombinedValues(values={})
 
 
 class MqttSimulationConnector(Connector[CombinedValues, CombinedValues]):
     def __init__(
         self,
         simulation: "Simulation",
         mqtt_client: Client,
-        topic_prefix: str,
+        amcs_topic_prefix: str,
```
aljevandam: perhaps these arguments can all just be plain topic_prefix if no other specification is needed

aljevandam (approved):

zero-thrs-control/src/thrs/cli/simulation_controls.py
hmvp: changed to devices_topic_prefix

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -76,7 +98,9 @@ def __init__(self, cls: type[M], topic_prefix: str):
         self._topic = topic_prefix
         self._value = None
 
-    def split_to_topics(self, model: M) -> dict[str, str]:
+    def split_to_topics(
```
aljevandam: computed_values should be removed here as argument?

hmvp: fixed

PR #382, opened 2026-06-23
Prevent overfill in dhw

aljevandam:


zero-thrs-control/src/thrs/control/modules/dhw.py
```diff
@@ -438,14 +440,47 @@ def __call__(self, sensor_values: DhwSensorValues, parameters: DhwParameters):
...
+                            else (
+                                TankState.NEEDS_FILL
+                                if tank.fillable(parameters)
+                                else (TankState.STANDBY)
+                            )
+                        )
+                    )
+                )
+            )
+        )
```
hmvp:
```python
    if tank is self._filling_tank:
        return TankState.FILLING
    if tank is self._boosting_tank:
        return TankState.BOOSTING
    if tank is self._tank_in_use:
        return TankState.IN_USE
    if tank.disabled:
        return TankState.DISABLED
    if tank.boostable(parameters):
        return TankState.NEEDS_BOOST
    if tank.fillable(parameters):
        return TankState.NEEDS_FILL
    return TankState.STANDBY
```

zero-thrs-control/src/thrs/control/modules/dhw.py
```diff
@@ -226,56 +229,44 @@ def boost(self, time: Callable[[], datetime]):
...
 
     def fillable(self, parameters: DhwParameters) -> bool:
-        return not self.disabled and not self.full(parameters)
+        return not self._disabled and not self._full and self.outlet_closed()
 
     def use(self, time: Callable[[], datetime]):
-        self._in_use = True
-        if self._empty_valve.setpoint.value != Valve.OPEN:
-            self._empty_valve.setpoint = Stamped(value=Valve.OPEN, timestamp=time())
+        self._full = False
```
hmvp, line 258: Is there a situation possible where a tank is "used" for a short while without actually draining much? Like could it go back into filling mode and then overfill because it is effectively already full...

hmvp (commented):
LGTM but I don't fully grasp the control logic yet...

zero-thrs-control/src/thrs/control/modules/dhw.py
aljevandam: Could be, possiby. But then it should still not overfill with this filling logic based on time to fill. 

I'm sure we will run into more issues - for now this solves the overfiling problem. 

In the future we might add a level>setpoint condition

zero-thrs-control/src/thrs/control/modules/dhw.py
```diff
@@ -226,56 +229,44 @@ def boost(self, time: Callable[[], datetime]):
...
-        if self._fill_valve.setpoint.value != Valve.OPEN:
-            self._fill_valve.setpoint = Stamped(value=Valve.OPEN, timestamp=time())
+        if self._inlet.setpoint.value != Valve.OPEN:
+            self._inlet.setpoint = Stamped(value=Valve.OPEN, timestamp=time())
 
     def stop_filling(self, time: Callable[[], datetime]):
-        self._filling = False
-        if self._fill_valve.setpoint.value != Valve.CLOSED:
-            self._fill_valve.setpoint = Stamped(value=Valve.CLOSED, timestamp=time())
+        self._full = True
```
mirdesign, line 250: Stop fill is always related to full?
Could have seperate 'def set_full(self)' ?

zero-thrs-control/src/thrs/control/modules/dhw.py
```diff
@@ -226,56 +229,44 @@ def boost(self, time: Callable[[], datetime]):
...
+        if self._inlet.setpoint.value != Valve.CLOSED:
+            self._inlet.setpoint = Stamped(value=Valve.CLOSED, timestamp=time())
 
     def fillable(self, parameters: DhwParameters) -> bool:
-        return not self.disabled and not self.full(parameters)
+        return not self._disabled and not self._full and self.outlet_closed()
```
mirdesign: Style idea;

` (not self._disabled) and (not self._full) and self.outlet_closed()`

or even

```
is_enabled = not self._disabled
is_empty = not self._full
is_outlet_closed = self.outlet_closed()

return is_enabled and is_empty and is_outlet_closed
```

zero-thrs-control/src/thrs/control/modules/dhw.py
```diff
@@ -226,56 +229,44 @@ def boost(self, time: Callable[[], datetime]):
...
-        elif self.boostable(parameters):
-            return TankState.NEEDS_BOOST
-        elif self.fillable(parameters):
-            return TankState.NEEDS_FILL
-        else:  # standby
-            return TankState.STANDBY
+    def outlet_closed(self) -> bool:
+        return (
+            self._outlet_position < (Valve.CLOSED + 0.001)
+            if self._outlet_position is not None
```
mirdesign: Style idea; 
```

if self._outlet_position is None:
    return False

return self._outlet_position < (Valve.CLOSED + 0.001)
```

zero-thrs-control/src/thrs/control/modules/dhw.py
```diff
@@ -339,27 +337,31 @@ def _select_tank_in_use(self, parameters: DhwParameters):
...
     ):
-        if self._filling_tank is not None and self._filling_tank.full(parameters):
-            self._filling_tank.stop_filling(self._time)
-
-            if self._filling_valves_closed(
+        if self._filling_tank:
+            if not self._filling_tank.full:
+                time_to_fill = self.time_to_fill(sensor_values, parameters)
+                if (
+                    time_to_fill and time_to_fill < 90
```
mirdesign, line 350: make 90 a CONST? you will use it on many other places

mirdesign (approved):

PR #387, opened 2026-06-24
Run simulations in separate pods

mirdesign:


docker-compose.yml
```diff
@@ -1,3 +1,20 @@
...
+    context: ./zero-thrs-control
+    dockerfile: Dockerfile
+    target: simulation
+  profiles:
+    - thrs
+  environment: &thrs_env
+    MQTT_HOST: vernemq
+    MQTT_PORT: 1883
+    MQTT_TOPIC_PREFIX: thrs
+    MQTT_CONTROL_TOPIC_SUFFIX: Command
```
hmvp: You probably need to update this to what is now line 350-356 and change that env to:
`<<: *thrs_env`

docker-compose.yml
```diff
@@ -341,18 +358,29 @@ services:
...
+    <<: *thrs_sim_template
+    command: ["consumers"]
+
+  thrs-thruster-simulation-th-high_temperature:
+    <<: *thrs_sim_template
+    command: ["high_temperature"]
+
+  thrs-thruster-simulation-th-boilers:
+    <<: *thrs_sim_template
+    command: ["boilers"]
```
hmvp: dhw

docker-compose.yml
```diff
@@ -341,18 +358,29 @@ services:
...
+    <<: *thrs_sim_template
+    command: ["pcm"]
+
+  thrs-thruster-simulation-th-consumers:
+    <<: *thrs_sim_template
+    command: ["consumers"]
+
+  thrs-thruster-simulation-th-high_temperature:
+    <<: *thrs_sim_template
+    command: ["high_temperature"]
```
hmvp: I am not sure if this one can be started with the other, so you might want to add profiles her or handle that in some way

zero-thrs-control/charts/zero-thrs-control/values.yaml
```diff
@@ -33,6 +33,7 @@ simulation:
     # Overrides the image tag whose default is the chart appVersion.
     tag: ""
   resources: {}
+  types: ["thrusters", "pvt", "pcm", "consumers", "high_temperature", "dhw"]
```
hmvp: I am not sure if high temp needs to run on the boat, it seemed to me that it was a attempt to run multiple control modules against a single special (cosimulated) fmu

aljevandam: The high temperature fmu is a combination of thrusters, pvt, pcm and consumers (combined in the modelica model, not cosimulated). It is the most representative we have at this point in that it requires multiple control module to run it. 

We will work towards simulating the whole system with a cosimulation of high_temp, adsorption, dhw, dc and drives. 

long story short - we want it, although whether we want to run any simulation on the boat is a different question.. (I'd say no)

zero-thrs-control/Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN apt-get update && \
 RUN fmpy compile /app/thrs/simulation/models/Thruster_Module_V19.fmu
```
aljevandam, line 43: do we need to compile thrusters here? 

if so, do we need to compile all fmus (or the right one?)

zero-thrs-control/README.md
```diff
@@ -23,6 +23,10 @@ CFLAGS="-std=gnu17" poetry run fmpy compile <path-to-fmu-file>
 
 ## Simulator
 
+Simulator deploy on Singel consist of multiple (6 at this point) `Deployments`, one for each `type` of simulation (e.g., thrusters, PVT).
+This is done, because FMU requires control signals to stay alive for a long time. Simulations tend to be used one per time, which will cause the others to crash at some point due those do not receive their control signals.
```
aljevandam: I find this explanation a bit strange and can't quite follow. 

Don't we just want them deployed so they can run independently and are ready to go? (eg when selecting a different simulation in the ui) 

Do we even need to 'justify' this or rather just explain what's happening. 'Deploys 6 simulations, intended to be used one at a time'.

aljevandam (commented):
sorry, forgot to submit this last week

zero-thrs-control/Dockerfile
florissmit10: Valid point. This has quite a big impact on the build time of the container.

Maybe we can compile the in parallel in a previous CI Step?

hmvp: Maybe we don't if we commit the compiled versions to git, This does mean that when there are new versions of FMUs its more work.. But it saves a lot in build time

florissmit10: Yes,
Commit compiled versions to git, and maybe make tests to see if they work on the CI?

Boudewijn26 (approved):

PR #389, opened 2026-06-24
Remove MqttSimulatorConnector

hmvp:
The end result is still not the prettiest, but we are getting there...

This also has the effect of splitting simulation outputs to dedicated topics, which is desirable, but can also be changed by adding layers in the SimulationValues classes used

We may decide on another name then controller_values...


zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -810,25 +809,30 @@ async def run(self, mode: Modes):
                 controller_topic_prefix=self._controller_topic_prefix,
                 sensor_values_clss=modules.sensor_values_clss,
                 control_values_clss=modules.control_values_clss,
+                controller_values_clss={},  # Nothing yet, but we want to send controller values here at some point
```
aljevandam, line 812: how about control_state_values (including state machine mode, controllers, tank states, etc) as a name?  So a ControlStateValues(ThrsModel) and a topic control-state/

possibly even put control parameters in the same topic and follow a similar /Command structure as in AMCS for setting parameters?

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -810,25 +809,30 @@ async def run(self, mode: Modes):
                 controller_topic_prefix=self._controller_topic_prefix,
```
aljevandam, line 809: so this would then be control_state_prefix

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -810,25 +809,30 @@ async def run(self, mode: Modes):
...
                 sensor_values_clss=modules.sensor_values_clss,
                 control_values_clss=modules.control_values_clss,
+                controller_values_clss={},  # Nothing yet, but we want to send controller values here at some point
                 control_topic_suffix=self._control_topic_suffix,
```
aljevandam, line 813: would devices_topic_control_suffix be more clear?

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -810,25 +809,30 @@ async def run(self, mode: Modes):
...
                 mqtt_client=self._sensor_client,
                 devices_topic_prefix=self._devices_topic_prefix,
-                simulation_topic_prefix=self._simulation_topic_prefix,
-                sensor_values_clss=modules.sensor_values_clss,
-                simulation_outputs_cls=modules.simulation_outputs_cls,
+                controller_topic_prefix=self._simulation_topic_prefix,
+                sensor_values_clss={},  # It actually does not listen to mqtt for these but gets them directly from the runner
+                control_values_clss=modules.sensor_values_clss,
+                controller_values_clss={mode: modules.simulation_outputs_cls},
+                sensor_topic_suffix=self._control_topic_suffix,
```
aljevandam, line 822: why a sensor topic suffix?

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -170,32 +169,36 @@ def result(self) -> CombinedValues | None:
...
     ):
         self._mqtt_client = mqtt_client
         self._sensor_values_mqtt_mapping = ModuleMqttMapping(
-            sensor_values_clss, devices_topic_prefix
+            sensor_values_clss, devices_topic_prefix, sensor_topic_suffix
         )
         self._control_values_mqtt_mapping = ModuleMqttMapping(
             control_values_clss, devices_topic_prefix, control_topic_suffix
         )
         self._computed_values_mqtt_mapping = ModuleMqttMapping(
```
aljevandam, line 196: I would propose calling this enriched_sensor_values_mqtt_mapping or computed_sensor_values_mqtt_mapping (other classes can also have computed values), and sending it to a topic with similar prefix (enriched_sensor_values or something like that)

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -170,32 +169,36 @@ def result(self) -> CombinedValues | None:
...
         sensor_values_clss: ModuleClassMap,
         control_values_clss: ModuleClassMap,
+        controller_values_clss: ModuleClassMap,
         control_topic_suffix: str | None = None,
+        sensor_topic_suffix: str | None = None,
     ):
         self._mqtt_client = mqtt_client
         self._sensor_values_mqtt_mapping = ModuleMqttMapping(
-            sensor_values_clss, devices_topic_prefix
+            sensor_values_clss, devices_topic_prefix, sensor_topic_suffix
```
aljevandam, line 191: or is the sensor topic suffix intenxed to be left None (at least on board?)

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -246,66 +255,16 @@ async def run(self):
         finally:
             self._running = False
 
-    async def transceive(self, control_values: CombinedValues) -> CombinedValues:
+    async def transceive(
```
aljevandam, line 258: I'd still  opt for calling 'send' and 'get' methods from the runner, not in a single tranceive - so that each published en collected object is for the same iteration

zero-thrs-control/src/thrs/orchestration/runner.py
```diff
@@ -68,50 +64,54 @@ def simulation(self):
...
-            connector,
-            module.control(initial_control_parameters, connector.time),  # type:ignore
-            module.alarms(),  # type:ignore
-        )
-
     async def run(self, n_ticks: int) -> None:
         for _ in range(n_ticks):
             sensor_values = await self._control_connector.transceive(
-                self._control_values  # type:ignore
+                self._control_values,  # type: ignore
```
aljevandam, line 90: the tranceive isnt just a matter of  'style' right? 

If on iteration n  I 'tranceive' then I'll have sensor values at time n, and send control values based on sensors of time  (n-1)

the control will be one step behind. 


if on iteration n I do 

get sensor values
run control 
send control values

I have sensor values at time n, and control values based on time n (which is what I would want?) 

or am i missing something

zero-thrs-control/src/thrs/cli/simulation_controls.py
hmvp: Yes, I was think about something like (with every key on a seperate topic)):
```python
class DhwControllerValues(ThrsValues):
    control_mode: DhwControlMode
    parameters: DhwParameters
    dhw_tanks_controller: controllers.TankControllerValues
    dhw_pump_flow_controller: controllers.PidControllerValues
    dhw_pump_temperature_controller: controllers.PidControllerValues
    dhw_drives_flow_controller: controllers.PidControllerValues
    dhw_dc_flow_controller: controllers.PidControllerValues
```

zero-thrs-control/src/thrs/cli/simulation_controls.py
hmvp: Broader than control_state, see above

zero-thrs-control/src/thrs/cli/simulation_controls.py
hmvp: probably

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: Currently this is only for the computed values as they come from SensorValues, but we can bikeshed the name

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: for the control modules, yes. sensor_topci_suffix is only for the sim... I might make some sense to incorporate the suffix into the Thrsvalues class, but we don't have anything for that atm

zero-thrs-control/src/thrs/orchestration/connector.py
hmvp: Agreed

zero-thrs-control/src/thrs/orchestration/runner.py
hmvp: You are correct, the fact it is one method is purely based on that is always was. (And called tick)

aljevandam (approved):

PR #391, opened 2026-06-26
Allow thrs with only control modules to run on boat

hmvp:


aljevandam (approved):

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -192,9 +192,25 @@
...
+    "boat", "thrusters", "pvt", "pcm", "consumers", "high_temperature", "dhw"
+]
 
-MODES: dict[Modes, tuple[str, CombinedModule]] = {
+MODES: dict[Modes, tuple[str | None, CombinedModule]] = {
+    "boat": (
+        None,
+        CombinedModule(
+            {
+                "thrusters": THRUSTERS_MODULE_DESCRIPTION,
```
aljevandam, line 204: the boat mode would also have adsorption, dc and drives (but given that things will move around we can add those later I guess)

PR #395, opened 2026-06-29
Move test models to test code

hmvp:


aljevandam (approved):
This is not necessarily test code though - I'd say it's for analysis. They might be better of in analysis/helpers of analysis_utils.py?

hmvp:
> This is not necessarily test code though - I'd say it's for analysis. They might be better of in analysis/helpers of analysis_utils.py?

Yeah, we should do something like that. For this code I expect part of this to be gone later so I wanted to move this out of the production code for now...

PR #396, opened 2026-06-29
Add missing env vars for deployment

hmvp:
Currently using the old values untill we setteled on the right values

### DIFF
--- a/zero-thrs-control/charts/zero-thrs-control/templates/configmap.yaml
+++ b/zero-thrs-control/charts/zero-thrs-control/templates/configmap.yaml
@@ -7,6 +7,8 @@ metadata:
 data:
   MQTT_HOST: {{ .Values.mqtt.host }}
   MQTT_PORT: {{ .Values.mqtt.port | quote }}
-  MQTT_TOPIC_PREFIX: thrs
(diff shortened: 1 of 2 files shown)


florissmit10 (approved):

zero-thrs-control/charts/zero-thrs-control/templates/configmap.yaml
```diff
@@ -7,6 +7,8 @@ metadata:
 data:
   MQTT_HOST: {{ .Values.mqtt.host }}
   MQTT_PORT: {{ .Values.mqtt.port | quote }}
-  MQTT_TOPIC_PREFIX: thrs
+  MQTT_DEVICES_TOPIC_PREFIX: thrs
+  MQTT_CONTROLLER_TOPIC_PREFIX: thrs
```
florissmit10: Maybe you can add a value for these things in values.yaml ?

hmvp: Yeah, agreed  I went for speed instead of correctness, but this should come from values.yaml indeed

florissmit10 (approved):

PR #397, opened 2026-06-29
Add missing modules

aljevandam:


hmvp (approved):

hmvp:
You are missing the bits in the graphql folder to allow actual use by the ui of these modules...

PR #398, opened 2026-06-29
Add ControllerState to control 

hmvp:
And move Dhw Controller output to there

This PR is big! But is is basically splitting ControlValues into ControlValues and ControllerValues. The former is sent to AMCS devices to do actual control and the latter is basically internal state of the control module that is needed in the frontend and data platform


zero-ui/src/modules/thrapp/mimics/tooltips/partials/BoilerTankController.vue
```diff
@@ -1,28 +1,28 @@
...
 
 defineProps<{
-  controller: ModuleField<ControlComponentType.DhwTanksController>;
+  controller: ModuleField<ControllerValueComponentType.DhwTanksController>;
 }>();
 </script>
 
 <template>
-  <ControlValue :source="controller">
+  <ControllerValue :source="controller">
```
null-33: I believe this is missing an import?

null-33 (commented):
LGTM. We should probably make it consistent and rename everything to `sensorValues` and `controlValues` (in another ticket).

null-33 (approved):

zero-ui/src/modules/thrapp/mimics/tooltips/partials/BoilerTankController.vue
hmvp: Weird, that was not noticed by the IDE and not by running `pnpm run build`... Fixes

zero-thrs-control/src/thrs/control/modules/dc.py
```diff
@@ -359,7 +353,11 @@ def _control_groups(self, sensor_values: DcSensorValues):
             )
         )
 
-        self._update_group_control_values()
+        self._update_group_control_values(
```
aljevandam, line 356: I guess in the future we'll need to update the contrrollervalues here as well

zero-thrs-control/src/thrs/control/modules/dhw.py
```diff
@@ -27,6 +28,49 @@
 from thrs.input_output.modules.dhw import DhwControlValues, DhwSensorValues
 
 
+class DhwControllerValues(ThrsValues):
+    dhw_tanks_controller: Annotated[
+        controllers.TanksControllerValues,
+        component_meta(
+            component_type="tank_controller",
+            included_in_fmu=False,
+            topic="controller/dhw/dhw-tanks-controller",
```
aljevandam: aren't the topics generated automatically just as in control values. I was expecting the prefix to be different in the connector, and otherwise no change

zero-thrs-control/tests/modules/thrusters/test_io.py
```diff
@@ -50,12 +52,14 @@ def test_yard_tags():
...
-    control_values = control.initial()
+def test_set_module_temperature(
+    control: ThrustersControl, simulation: ThrustersSimulation
+):
+    control_values, _ = control.initial()
 
-    simulation._simulation_inputs.thrusters_thruster_aft.heat_flow.value = 0
-    simulation._simulation_inputs.thrusters_thruster_fwd.heat_flow.value = 0
-    simulation._simulation_inputs.thrusters_pcm_supply.temperature.value = 60
+    simulation._simulation_inputs.thrusters_thruster_aft.heat_flow.value = 0  # type: ignore
```
aljevandam, line 60: why is the type ignore needed now?

aljevandam (approved):
I reviewed the Python part. 

I think very good - but I'd like to make a case for

- ControlState instead of ControllerValues, and adding the mode in there as well instead of having separate classes and topics for those 
- Perhaps it's worth reinstating the ControlResult class to hold these classes instead of having control return tuples

zero-thrs-control/src/thrs/control/modules/dhw.py
hmvp: Yeah this custom topic can be removed

zero-thrs-control/tests/modules/thrusters/test_io.py
hmvp: might be a reformatting thing, will check

zero-thrs-control/src/thrs/control/modules/dc.py
hmvp: Yeah this might change in the future

hmvp:
> I reviewed the Python part.
> 
> I think very good - but I'd like to make a case for
> 
>     * ControlState instead of ControllerValues, and adding the mode in there as well instead of having separate classes and topics for those
That is not a bad idea!

Btw I was planning on moving the mode, but the PR is already quite big so thats a separate one. Same for parameters (outgoing)
> 
>     * Perhaps it's worth reinstating the ControlResult class to hold these classes instead of having control return tuples
Yeah I don't like the tuples either, good point
EDIT: Hmm tuples work better with argument unpacking, with a ControlResult class this might get uglier...

zero-thrs-control/tests/modules/thrusters/test_io.py
hmvp: Ah because the argument type is given for the fixture so typechecking is done and `value` is of type `Stamped |SkipJsonSchema[...]` and the latter is something weird

PR #399, opened 2026-06-29
Remove simulation classes from CombinedModule

aljevandam:


aljevandam:
Still some work to do on the typing and use of Simulation class I suppose, and perhaps we want restrict the interface of Module and Simulation to only take CombinedValues

zero-thrs-control/src/thrs/orchestration/runner.py
```diff
@@ -1,13 +1,41 @@
...
+    tick_duration: timedelta = timedelta(seconds=1)
+
+    def simulation(self) -> Simulation:
+        return Simulation(
+            self.sensor_values_cls,
+            self.simulation_outputs_cls,
+            Fmu(self.fmu_path),
+            self.simulation_inputs,
+            self.start_time,
+            self.tick_duration,
```
hmvp: This is probably a leftover from merging and can be removed (it isn't used)

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -296,102 +297,163 @@
...
+@dataclass
+class Mode:
+    name: Modes
+    control_module: CombinedModule
+    simulation_getter: Callable[[], Simulation | None]
+
+
+MODE_CONFIGS: dict[str, Mode] = {
+    "thrusters": Mode(
+        name="thrusters",
```
Boudewijn26: You could deduplicate this by literalling a list instead of a dict and mapping that list to a dict

Boudewijn26 (approved):

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -922,51 +992,44 @@ async def run(self, mode: Modes):
...
-                    modules.sensor_values_clss,
-                    modules.simulation_outputs_cls,
-                    fmu,
-                    simulation_inputs,
-                    datetime.now(),
-                    tick_duration,
-                )
-                time_func = simulation.time
+        mode = MODE_CONFIGS[mode_name]
+        simulation_inputs = SIMULATION_INPUTS[mode_name]
```
hmvp: This gives an error if there are no sim inputs (mode=boat)

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -976,50 +1039,57 @@ async def run(self, mode: Modes):
...
-                        SimulationInputMessage(
-                            inputs=cast(ThrustersSimulationInputs, simulation_inputs)
-                        ),
-                    )
+                await context.send(
+                    self._simulation_topic_prefix,
+                    SimulationInputMessage(
+                        inputs=cast(ThrustersSimulationInputs, simulation_inputs)
+                    ),
+                )
```
hmvp: Shouldn't this be conditional on if there is a simulation?

zero-thrs-control/src/thrs/cli/simulation_controls.py
```diff
@@ -296,102 +297,163 @@
...
             {"pcm": PCM_MODULE_DESCRIPTION},
+        ),
+        simulation_getter=lambda: Simulation(
             PcmSimulationInputs,
             PcmSimulationOutputs,
+            Fmu(pcm_path),
+            SIMULATION_INPUTS["pcm"],
+            datetime.now(),
+            timedelta(seconds=1),
         ),
```
hmvp, line 366: I don't really like that we need to specify al the time stuff here... That feels like its it should come from the runtime

maybe something like (or a similar helper function):
```diff
@@ -313,13 +314,12 @@ MODE_CONFIGS: dict[str, Mode] = {
                 "thrusters": THRUSTERS_MODULE_DESCRIPTION,
             },
         ),
-        simulation_getter=lambda: Simulation(
+        simulation_getter=partial(
+            Simulation,
             {"thrusters": ThrustersSensorValues},
             ThrustersSimulationOutputs,
             Fmu(thrusters_path),
             SIMULATION_INPUTS["thrusters"],
-            datetime.now(),
-            timedelta(seconds=1),
         ),
     ),
     "pvt": Mode(
```

and then:

```diff
@@ -995,7 +987,7 @@ class SimulationControls:
         mode = MODE_CONFIGS[mode_name]
         simulation_inputs = SIMULATION_INPUTS[mode_name]
         control_module = mode.control_module
-        simulation = mode.simulation_getter()
+        simulation = mode.simulation_getter(datetime.now(), timedelta(seconds=1))
 
         simulation_context = simulation if simulation else nullcontext()
         with simulation_context:
```

aljevandam: The point is to reinstantiate the FMU so I need to 'generate' it. Requires a different solution

hmvp: Yeah the above does that, it just generates the getter based on the Constructor instead of a lambda and moves the time arguments to the getter

```diff
@@ -313,13 +314,12 @@ MODE_CONFIGS: dict[str, Mode] = {
                 "thrusters": THRUSTERS_MODULE_DESCRIPTION,
             },
         ),
-        simulation_getter=lambda: Simulation(
+        simulation_getter=lambda time_fn, tick_duration: Simulation(
             {"thrusters": ThrustersSensorValues},
             ThrustersSimulationOutputs,
             Fmu(thrusters_path),
             SIMULATION_INPUTS["thrusters"],
-            datetime.now(),
-            timedelta(seconds=1),
+            time_fn,
+            tick_duration,
         ),
     ),
     "pvt": Mode(
```
does the same but without `partial`

aljevandam: hm yeah I might have gotten confused. I tried with the partial but it didn't work - I don't think the Fmu gets reinstantiated when you call the partial with the time arguments?

the lambda alternative would work I suppose

PR #401, opened 2026-06-30
Add fiber optics MQTT adapter

Boudewijn26:
I picked Rust for this since it can become quite high bandwidth (at 100 Hz)


.github/workflows/ci_fiber_optics.yaml
```diff
@@ -0,0 +1,86 @@
...
+        working-directory: zero-fiber-optics
+
+      - name: Run clippy
+        run: cargo clippy --all --all-features --locked
+        working-directory: zero-fiber-optics
+
+      - name: Determine image tag
+        id: image_tag
+        run: |
+          VERSION=$(grep '^version = ' Cargo.toml | head -1 | sed -E 's/version = "(.*)"/\1/')
```
florissmit10: Wont this run in line with the rest of the codebase? Release based on github release and tags?

zero-fiber-optics/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,96 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
+        fsGroupChangePolicy: Always
+        sysctls: []
+        supplementalGroups: []
+        fsGroup: 1001
+      containers:
+        - name: fiber-optics
+          securityContext:
```
florissmit10, line 36: This can be a lot shorter because many things here are the default:

```
          securityContext:
              allowPrivilegeEscalation: false
              capabilities:
                drop:
                - ALL
              readOnlyRootFilesystem: true
              runAsNonRoot: true
```

zero-fiber-optics/src/asyncapi.rs
```diff
@@ -0,0 +1,286 @@
...
+    MessageRef, Operation, OperationAction, Schema, SchemaObject,
+};
+use crate::config::UdpChannels;
+use crate::layout::{topic_segment, TopicMap};
+
+/// Dynamically build an AsyncAPI 3.0.0 schema from the loaded configuration.
+///
+/// Each port channel publishes one packet object under `{prefix}/{channel_name}`.
+/// Object properties are keyed by resolved topic segment (override first, kebab-case fallback).
+pub fn build_schema(config: &UdpChannels, prefix: &str, topic_map: &TopicMap) -> AsyncApiSpec {
```
florissmit10, line 12: Is the `AsyncApiSpec` used for running the application? If so, how? if not, what is the usecase? It looks like a powerful standard

zero-fiber-optics/example/config.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version='1.0'?>
+<UDPChannels ip='0.0.0.0'>
```
florissmit10, line 2: Im assuming the source provides the full/real config file. Did they just make this spec up as they went, or is it a standard of some sort?

Boudewijn26: Normally you'd get an XML based on the configuration of the Bravo systems processor (so it's an artifact from running the processing). It's their own standard

zero-fiber-optics/src/asyncapi.rs
Boudewijn26: It's not used to run the application, but I want to move towards having **everything** described by asyncapi. We can then use that to do data processing etc

.github/workflows/ci_fiber_optics.yaml
Boudewijn26: Yes, it should

florissmit10:
If we want arm images, I think you need to run the different platforms on a different cpu architecture. I believe I did this for the powerhub. 
2 jobs
- 1 that builds the different architectures in paralell and pushes the layers
- 1 that publishes the final manifest based on the images.

https://github.com/foundation-zero/power-hub/blob/main/.github/workflows/docker.yml

If not it would be better to remove the arm build target

.github/workflows/ci_fiber_optics.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+      - name: Determine image tag
+        id: image_tag
+        run: |
+          if [ "${{ github.event_name }}" = "release" ]; then
+            VERSION="${{ github.event.release.tag_name }}"
+          elif [ "${{ github.ref_type }}" = "tag" ]; then
+            VERSION="${{ github.ref_name }}"
+          else
+            VERSION=$(grep '^version = ' Cargo.toml | head -1 | sed -E 's/version = "(.*)"/\1/')
```
florissmit10: I think in this case everyones feature branch will get pushed to the version in the cargo toml right? Maybe just add the git sha in that case?

florissmit10 (approved):
Looking good!
Rust <3

Boudewijn26:
> If we want arm images, I think you need to run the different platforms on a different cpu architecture. I believe I did this for the powerhub. 2 jobs
> 
> * 1 that builds the different architectures in paralell and pushes the layers
> * 1 that publishes the final manifest based on the images.
> 
> https://github.com/foundation-zero/power-hub/blob/main/.github/workflows/docker.yml
> 
> If not it would be better to remove the arm build target

I've now gone for an actual cross build. With rust there are several flavors:
* Just use `cross` (which uses docker under the hood)
* Cross build on native platform (what I've gone for)
* In Docker emulate target platform and build on there
* Run Actions on target platform and merge later

I think what I've done is the simplest and fastest. Emulation slows down the build by quite a lot. `cross` requires a separate build step to copy the resulting binary into a docker image. Merging complicates the github actions by quite a lot

PR #408, opened 2026-07-01
Add SailSystem topic to vector

florissmit10:
ZERO-1659


Boudewijn26 (approved):
Technically we should do kebab-case. That is what https://github.com/foundation-zero/zero-sail-systems-mqtt-adapter/pull/8 does

florissmit10:
There is illegal JSON in the messages, so for now I will wait before merging this

Boudewijn26:
No JSON is illegal 😉

PR #410, opened 2026-07-01
Remove FMU compile from docker image

hmvp:
Also adds some just file commands to (cross) compile the fmus after updating

This assumes that we run the just command to compile for linux locally whenever we update or add fmu files


zero-thrs-control/justfile
```diff
@@ -29,20 +29,21 @@ test_io *args:
...
-    just compile_fmu src/thrs/simulation/models/DC_Module.fmu
-    just compile_fmu src/thrs/simulation/models/DHW_Module.fmu
-    just compile_fmu src/thrs/simulation/models/Drives_Module.fmu
-    just compile_fmu src/thrs/simulation/models/HT_SystemV14.fmu
-    just compile_fmu src/thrs/simulation/models/PCM_ModuleV8.fmu
-    just compile_fmu src/thrs/simulation/models/PVT_ModuleV16.fmu
-    just compile_fmu src/thrs/simulation/models/Thruster_Module_V19.fmu
+[parallel]
+compile_fmus: (compile_fmu "src/thrs/simulation/models/XRGTestModel/FMUInterfaceTester_MECS_regular.fmu") (compile_fmu "src/thrs/simulation/models/Adsorption_Module.fmu") (compile_fmu "src/thrs/simulation/models/Consumer_ModuleV9.fmu") (compile_fmu "src/thrs/simulation/models/DC_Module.fmu") (compile_fmu "src/thrs/simulation/models/DHW_Module.fmu") (compile_fmu "src/thrs/simulation/models/Drives_Module.fmu") (compile_fmu "src/thrs/simulation/models/HT_SystemV14.fmu") (compile_fmu "src/thrs/simulation/models/PCM_ModuleV8.fmu") (compile_fmu "src/thrs/simulation/models/PVT_ModuleV16.fmu") (compile_fmu "src/thrs/simulation/models/Thruster_Module_V19.fmu")
```
aljevandam: is there a way to use the paths from fmu_paths.py here?

aljevandam (commented):
Would we then also want to  remove the compiling step from the CI so that when the FMU's don't run on linux it fails?

zero-thrs-control/justfile
hmvp: Probably not (or not without a lot of complexity)

hmvp:
> Would we then also want to remove the compiling step from the CI so that when the FMU's don't run on linux it fails?

Ah yes that should be possible

aljevandam (approved):

PR #411, opened 2026-07-01
Add dummy topic overrides where needed

aljevandam:


zero-thrs-control/src/thrs/input_output/modules/dc.py
```diff
@@ -232,49 +232,73 @@ class DcSensorValues(ThrsValues):
     dc_brightloop_aft1: Annotated[
         sensor.Brightloop,
         component_meta(
-            yard_tag="45002076", component_type="brightloop", included_in_fmu=False
+            yard_tag="45002076",
+            component_type="brightloop",
+            included_in_fmu=False,
+            topic_override="dummy-pms/brightloop-aft1-active",
```
hmvp, line 238: Why the dummy-pms prefix, don't we have the correct topics yet?

hmvp (approved):

zero-thrs-control/src/thrs/input_output/modules/dc.py
aljevandam: not that I could find out

PR #412, opened 2026-07-01
Separate executions

Boudewijn26:
This:

- Splits simulation controls into several components (all in runtime)
- Has separate runners for the different cases (control, simulation, lockstep)
- Creates case specific channels (Control, Simulation, Directives / ControlApi, SimulationApi, DirectivesApi)
- Uses those channels to centralize all MQTT communication


zero-thrs-control/src/thrs/graphql/messaging.py
Boudewijn26, line 1: With the move to Channels, these classes are doing very little. We could probably just replace them with their channels (but move something for a different PR)

zero-thrs-control/src/thrs/orchestration/simulation.py
```diff
@@ -28,10 +28,10 @@ class SimulationResult[
 
 
 class Simulation[
-    S: ThrsValues,
-    C: ThrsValues,
+    S,
+    C,
     I: SimulationInputs,
-    O: ThrsValues,
+    O: SimulationValues,
```
aljevandam, line 34: Wy is this a simulationvalues - simulation outputs don't need the dataframe stuff so I think it can be ThrsValues?

zero-thrs-control/src/thrs/runtime/runners.py
```diff
@@ -0,0 +1,170 @@
...
+
+class LockstepRunner[
+    S: CombinedValues,
+    P: CombinedValues,
+    M: CombinedValues,
+    I: SimulationInputs,
+    O: SimulationValues,
+    CS: CombinedValues,
+](Runner):
+    """Runs a module for a number of ticks."""
```
aljevandam: this description seems off. 'Runs a simulation and control module in tandem and optionally publishes results on mqtt' ?

zero-thrs-control/src/thrs/runtime/runners.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+
+            self._control_values, self._controller_state = self._control.control(
+                sim_result.sensor_values
+            )
+            alarms = self._alarms.check(
+                sim_result.sensor_values, self._control_values, self._control.parameters
+            )
+
+            await self._control_channels.send_controller_state(self._controller_state)
```
aljevandam: control values are published before the (simulation) tick, and the control state values after which seems strange. 

I would suggest publishing everything after

zero-thrs-control/src/thrs/runtime/runners.py
```diff
@@ -0,0 +1,170 @@
...
+            manual_modes = self._control_channels.get_automation_modes()
+            if manual_modes is not None:
+                self._control.update_automation_modes(manual_modes)
+            simulation_inputs = self._simulation_channels.get_simulation_inputs()
+            if simulation_inputs is not None:
+                self._simulation.update_simulation_inputs(simulation_inputs)
+
+            await self._control_channels.send_control_values(self._control_values)
+
+            sim_result = self._simulation.tick(self._control_values)
```
aljevandam: currently a bit of asymmetry/inconsistency here, with simulation returning a sim_result and control returning a tuple

zero-thrs-control/src/thrs/runtime/runners.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def _check_alarms(self, sensor_values):
+        alarms = self._alarms.check(
+            sensor_values, self._control_values, self._control.parameters
+        )
+        if alarms:
+            warnings.warn(f"Alarms detected: {alarms}")  # TODO: properly handle alarms
+
+
+class SimulationRunner[S: CombinedValues, C, I: SimulationInputs, O: SimulationValues](
```
aljevandam: O: ThrsValues?

zero-thrs-control/src/thrs/runtime/runners.py
```diff
@@ -0,0 +1,170 @@
...
+            parameters = self._control_channels.get_parameters()
+            if parameters is not None:
+                self._control.update_parameters(parameters)
+
+            manual_modes = self._control_channels.get_automation_modes()
+            if manual_modes is not None:
+                self._control.update_automation_modes(manual_modes)
+            simulation_inputs = self._simulation_channels.get_simulation_inputs()
+            if simulation_inputs is not None:
+                self._simulation.update_simulation_inputs(simulation_inputs)
```
aljevandam: does this always updated or is simulation_inputs only not None if it was changed?

zero-thrs-control/src/thrs/runtime/runtime.py
```diff
@@ -0,0 +1,195 @@
...
+            control = mode.control_module.control(
+                CombinedValues(parameters), datetime.now
+            )
+            runner = ControlRunner(
+                channels=control_channels,
+                control=control,
+                alarms=mode.control_module.alarms(),
+            )
+            yield Runtime(
+                runner, timedelta(seconds=1), connector_runs=[await connector.run()]
```
aljevandam: this one gets a default 1 sec while simulation gets a tick duration?

zero-thrs-control/src/thrs/runtime/runtime.py
```diff
@@ -0,0 +1,195 @@
...
+            directives_channels = DirectivesChannels(directive_connector, config)
+
+            directive_handling = DirectiveHandling(
+                directives_channels,
+                mode,
+                simulation.time,
+            )
+            yield Runtime(
+                runner,
+                simulation.tick_duration,
```
aljevandam: how does this work with simulation time, tick durations.. I assum the dick duration is default 1 second (in simulation time), and that the plkayback rate determines the ratio between a simulation second and a real second? 

So i'd expect an option here to set the playback rate en parhaps next to that the tick duration  (as with simulation and control)

zero-thrs-control/src/thrs/orchestration/simulation.py
Boudewijn26: The SimulationOutputs are using the same boundary classes as the SimulationInputs, so they do need to be dedataframeable. Also I prefer having these constraints be as specific as possible to prevent errors

zero-thrs-control/src/thrs/runtime/runners.py
Boudewijn26: True, but that wasn't introduced by this PR

zero-thrs-control/src/thrs/runtime/runners.py
Boudewijn26: Always updated (only `None` when SimulationInputs is incomplete)

zero-thrs-control/src/thrs/runtime/runtime.py
Boudewijn26: Yeah, should probably be a constant to signal its significance

zero-thrs-control/tests/cli/test_simulation_controls.py
```diff
@@ -1,510 +0,0 @@
-from asyncio import create_task, sleep
```
hmvp, line 1: I do think we should have some test like these that test the whole thing, they seem to be missing now

zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -0,0 +1,730 @@
...
+
+    async def _start(self):
+        for mapping in self._listeners:
+            for topic in mapping.subscribe_topics():
+                await self._mqtt_client.subscribe(topic, qos=1)
+
+    async def run(self) -> Coroutine[None, None, None]:
+        await self._start()
+        self._started = True
+        return self._listen()
```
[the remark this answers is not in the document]
Boudewijn26: This is intentional: we want to separate having done the subscriptions from running the loop

zero-thrs-control/src/thrs/runtime/runners/lockstep.py
```diff
@@ -0,0 +1,62 @@
...
+        for _ in range(n_ticks):
+            await self._sync_channels_state()
+
+            sim_result = await self.simulation_runner._execute_simulation_tick(
+                self._control_values
+            )
+
+            self._control_values = await self.control_runner._execute_control_tick(
+                sim_result.sensor_values
+            )
```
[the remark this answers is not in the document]
hmvp: This is intentional

zero-thrs-control/src/thrs/runtime/runners.py
hmvp: This is changed now

zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -0,0 +1,730 @@
...
+                    for topic in mapping.subscribe_topics()
+                    if message.topic.matches(topic)
+                ]
+                for mapping in matching_mappings:
+                    if not isinstance(message.payload, str | bytes):
+                        raise ValueError(
+                            f"Expected string or bytes, got {type(message.payload)}"
+                        )
+                    mapping.handle_message(message.topic.value, message.payload)
+
```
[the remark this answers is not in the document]
Boudewijn26: Nice find Copey, but the `clear_simulation_status` is just to remove the retain from previous executions. We don't want to clear the memory of anything that's running

zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -0,0 +1,730 @@
...
+    def subscribe_topics(self) -> set[str]:
+        return set(self._subscribe_topics.keys())
+
+    def handle_message(self, topic: str, json: str | bytes):
+        if topic not in self._topic_to_field:
+            return
+
+        self._builder.input(self._topic_to_field[topic], json)
+        self._update_event.set()
+
```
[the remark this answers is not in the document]
Boudewijn26: That is why empty payloads aren't delivered

zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -0,0 +1,730 @@
...
+    def handle_message(self, topic: str, json: str | bytes):
+        if topic == self._topic:
+            self._value = cast(M, self._adapter.validate_json(json))
+            if not self._future.done():
+                self._future.set_result(self._value)
+            self._update_event.set()
+            for hook in self._hooks:
+                result = hook(self._value)
+                if isawaitable(result):
+                    ensure_future(result)
```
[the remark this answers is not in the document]
Boudewijn26: See https://github.com/foundation-zero/zero/pull/412#discussion_r3542222710

zero-thrs-control/src/thrs/runtime/runtime.py
Boudewijn26: Good point, let's nail the durations in a different PR?

hmvp (approved):

mirdesign (approved):

PR #413, opened 2026-07-01
Add parameter output into controller values

hmvp:
This adds the parameters to ControllerState so that we publish the used parameters every control tick.


Boudewijn26 (commented):
Is there a plan to remove the `@property def parameters` from the controls? Is that scoped somewhere?

hmvp:
> Is there a plan to remove the `@property def parameters` from the controls? Is that scoped somewhere?

Ah yes that might be possible now (need to check)

zero-thrs-control/src/thrs/graphql/base.py
```diff
@@ -295,7 +295,7 @@ async def _mutation(
                 if parameters is None:
                     raise Exception("No parameters available to update")
                 setattr(parameters, name, value)
-                expect = mod.wait_for_parameters(
+                expect = mod.wait_for_controller_state(
                     lambda parameters: getattr(parameters, name) == value, timeout=2
```
Boudewijn26, line 299: At the very least naming here is wonky, but probably you'd expect `controller_state.parameters` to be there, but with it being a value in each `ControllerState` you have no guarantee the relevant control has parameters in the `ControllerState`

Do you consider it legal/valid for a `ControllerState` not to include a `parameters` field?

Boudewijn26 (approved):

zero-thrs-control/src/thrs/simulation/io_mapping.py
```diff
@@ -31,8 +29,9 @@ def flatten_model_values(
             {},
         )
     return {
-        fmu_key: get_value(model, component, field)
+        fmu_key: data.value if isinstance(data, Stamped) else data
         for (component, field), fmu_key in fmu_key_mapping.items()
+        if (data := getattr(getattr(model, component), field))
```
mirdesign: shouldn't this be `if (data := getattr(getattr(model, component), field)) is not None` ? Since now you're checking the value of 'data', having a 'false' value drop the mapping value.

zero-ui/scripts/extract-schema-values.ts
```diff
@@ -302,9 +290,14 @@ function processSensorField(
 }
 
 function processControllerField(
+  fieldName: string,
   fieldType: string,
   _directiveArgs: DirectiveArgs,
 ): ExtractedValue | null {
+  if (fieldName == "parameters") {
```
mirdesign: Do we really want this?
At least make "parameters" value a const somewhere? Related to something so it will not be forgotten during a future code change.

PR #417, opened 2026-07-03
Add example of how to specialize mqtt connector

Boudewijn26:
For consumers this is fully constrained by typing: 
1. The way to send e.g. control values is through the `ControlChannels`.
2. The only way to get an instance of `ControlChannels` is through `MqttConnector.run` (because the `__init__` and `register` are unavailable because you can't get a `_PrivateToken` from the outside)
3. The only way to call `MqttConnector.run` is by providing a `ChannelsDescription` which then binds the description into the `Channels`

If this was Rust we could consume the connector in `run` to prevent it from being called twice, but this isn't Rust.

### DIFF
--- a/zero-thrs-control/src/thrs/orchestration/connector.py
+++ b/zero-thrs-control/src/thrs/orchestration/connector.py
@@ -6,7 +6,13 @@ from typing import Awaitable, Callable, Protocol, Self
 from aiomqtt import Client
 from pydantic.fields import ComputedFieldInfo, FieldInfo
 
-from thrs.input_output.base import CombinedValues, ThrsValues, get_topic
+from thrs.input_output.base import (
+    CombinedValues,
+    SimulationInputs,
+    SimulationValues,
+    ThrsValues,
+    get_topic,
+)
 from thrs.input_output.model_builder import CombinedModelBuilder, PartialModelBuilder
 from thrs.orchestration.module import ModuleClassMap
 from thrs.utils.string import hyphenize
@@ -178,16 +184,11 @@ type Publisher[T] = Callable[[T], Awaitable[None]]
 
 class Channels[T](Protocol):
     @classmethod
-    def register(
+    def _register(
         cls,
-        private_token: "type[_PrivateToken]",
         connector: "MqttConnector",
         description: T,
     ) -> Self: ...
-
-
-class _PrivateToken:
-    pass
 
 
 @dataclass
@@ -195,26 +196,27 @@ class ControlChannelsDescription:
     sensor_values_clss: ModuleClassMap
     sensor_values_topic_prefix: str
     control_values_clss: ModuleClassMap
+    controller_state_clss: ModuleClassMap
     control_values_topic_prefix: str
     control_values_topic_suffix: str | None = None
-    # parameters, controller values etc.
 
 
 class ControlChannels(Channels[ControlChannelsDescription]):
     def __init__(
         self,
-        private_token: type[_PrivateToken],
         sensor_values_mapping: ModuleMqttMapping,
         send_control_values: Publisher[CombinedValues],
+        send_controller_state: Publisher[CombinedValues],
         # etc
     ):
+        """Don't ever call this directly, use `MqttController#run` instead"""
         self.get_sensor_values = sensor_values_mapping.result
         self.send_control_values = send_control_values
+        self.send_controller_state = send_controller_state
 
     @classmethod
-    def register(
+    def _register(
         cls,
-        _private_token: type[_PrivateToken],
         connector: "MqttConnector",
         description: ControlChannelsDescription,
     ) -> "ControlChannels":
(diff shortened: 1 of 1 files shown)


zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -177,6 +173,115 @@ async def run(self): ...
...
+    control_values_topic_suffix: str | None = None
+    # parameters, controller values etc.
+
+
+class ControlChannels(Channels[ControlChannelsDescription]):
+    def __init__(
+        self,
+        private_token: type[_PrivateToken],
+        sensor_values_mapping: ModuleMqttMapping,
+        send_control_values: Publisher[CombinedValues],
```
Boudewijn26, line 203: It could be argued that it'd be nicest to have both of these just be `Callable`s. Now only `Publisher` is. That would also lean more towards dataclassification

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -203,23 +308,39 @@ def __init__(
             controller_state_clss, controller_topic_prefix
         )
         self._running = False
+        self._listeners: list[MqttReceiveMapping[object]] = []
+
+    def register_listener(self, receiver: MqttReceiveMapping[object]) -> None:
```
hmvp: I don't like that you expose the Mqtt*Mappings in the api.. I consider them to be an internal implementation detail

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -177,6 +173,115 @@ async def run(self): ...
...
+    ):
+        self.get_sensor_values = sensor_values_mapping.result
+        self.send_control_values = send_control_values
+
+    @classmethod
+    def register(
+        cls,
+        _private_token: type[_PrivateToken],
+        connector: "MqttConnector",
+        description: ControlChannelsDescription,
```
hmvp, line 216: I am not necessary a fan of using dataclasses just to save on a few arguments. Its fine if the dataclass is filled somewhere far away, but this feels like a case of the caller needing to create a dataclass and the callee (this function) needing to directly unpack it again. Just use more arguments in that case and save all the generic hassle (especially since both ChannelDescriptions are identical)

hmvp (commented):
This feels like a lot of extra code for almost no new functionality/capabilities. You introduce a lot of generics and flexibility which IMHO is not going to be used because we will not have N new types of modules (besides control and simulation)

However this seems only part of the implementation so I might miss something here..

I do appreciate the effort though! It is interesting to see in which direction you want to take it..

zero-thrs-control/src/thrs/orchestration/connector.py
Boudewijn26: This method should only be used in `connector.py`, but isn't private to `MqttConnector` because it's used by the `Channels`

Boudewijn26:
@hmvp I'm not sure you read the PR description. The way you'd use this has no generics and the benefit of only needing to provide the arguments you need for the specific case whilst protecting you from forgetting one of the classes you need to provide.
```python
connector = MqttConnector(mqtt_client)
control_channels = await MqttConnector.run(ControlChannels, ControlChannelsDescription(
  sensor_values_clss=...,
  control_values_clss=...,
  # topic prefices etc
))
```

zero-thrs-control/src/thrs/orchestration/connector.py
Boudewijn26: The `ChannelDescription`s are just indicative, they would differ in `ControlChannelDescription` having the `ControllerState` and the `SimulationChannelDescription` having `SimulationOutputs` and `SimulationInputs` that's why the comments are there

Like shown in https://github.com/foundation-zero/zero/pull/417#issuecomment-4874918442 the caller wouldn't be calling this function (and they can't because they can't provide the `_PrivateToken`)

The generic is used to ensure the provided description matches the channels, since this is going through `MqttConnector#run` you'd have to cast values if they were arbitrary args and it'd be easy to miss some

Boudewijn26:
@hmvp I've cleaned up the example quite a bit. I hope this is clearer

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -177,97 +179,185 @@ async def run(self): ...
...
-            control_values_clss, devices_topic_prefix, control_topic_suffix
+
+
+@dataclass
+class SimulationChannelsDescription[I: SimulationInputs, O: SimulationValues]:
+    sensor_values_clss: ModuleClassMap
+    sensor_values_topic_prefix: str
+    control_values_clss: ModuleClassMap
+    simulation_inputs_cls: type[I]
+    simulation_outputs_cls: type[O]
```
Boudewijn26, line 248: Perhaps the best reason for doing something in this direction is that the SimulationInput and SimulationOutput can just be their actual classes without having to be wrapped into an arbitrary `ModuleClassMap`

zero-thrs-control/src/thrs/orchestration/connector.py
```diff
@@ -177,97 +179,185 @@ async def run(self): ...
...
+                    description.simulation_outputs_cls,
+                    description.controller_topic_prefix,
+                ),
+            ),
         )
-        self._running = False
 
-    async def _listen_to_sensors(self):
+
+class MqttConnector(Connector[CombinedValues, CombinedValues, CombinedValues]):
```
Boudewijn26, line 311: With this PR the `MqttConnector` is fully limited (at least from the outside) to just doing MQTT and how it does it is fully hidden

Any actual sending/receiving happens through the `Channels`. Which would pair well with `MqttConnector` living in `Runtime` and the `Channels` living in the `Runner`s

Boudewijn26:
Done in #412

PR #419, opened 2026-07-03
Update loads app mqtt topics

aljevandam:


zero-loads-app/src/loads/sensors/sail-system-mqtt-schema.txt
```diff
@@ -0,0 +1,9449 @@
+Project Name: 3094_SailPLC
```
Boudewijn26: If you take this line out, this is valid JSON (and thus you could rename the file)

zero-loads-app/src/loads/sensors/sail_system.py
```diff
@@ -493,112 +766,165 @@ class MizzenCheckstay(LoadsModel, ABC):
...
+            display_name="Checkstay SB Load Failure",
+            alarm_for="sb_load",
+        ),
+    ]
+    max_load_ps: Annotated[
+        MaxLoad,
+        Field(validation_alias="st_LoadPs/i_MaxLoadSetting", ge=0, le=2.7),
+        VariableMeta(
+            name="max-ps-load",
+            display_name="Checkstay PS Max Load",
```
Boudewijn26: I'm seeing a lot of portside being PT, but this is still PS. Are we switching or not?

zero-loads-app/src/loads/sensors/base.py
```diff
@@ -76,11 +76,11 @@ def _create_generator(cls, base_type: Any, meta: list[Any]) -> Generator:
 
     @staticmethod
     def extract_minimum(meta: list[Any]) -> float | None:
-        return next((cast(float, m.ge) for m in meta if isinstance(m, Ge)), None)
+        return next(cast(float, m.ge) for m in meta if isinstance(m, Ge))
```
Boudewijn26: This can never return `None`, right?

Boudewijn26 (approved):

zero-loads-app/src/loads/sensors/sail_system.py
aljevandam: good catch

PR #420, opened 2026-07-06
Fix timout during manual control

hmvp:


zero-thrs-control/src/thrs/control/switching.py
```diff
@@ -56,7 +56,7 @@ def control(
     ) -> tuple[ControlValues, ControllerState]:
         if self._mode == "manual":
             control_values, _ = self._manual_control.control(sensor_values)
-            return control_values, EmptyControllerState()  # type: ignore
+            return control_values, self._automatic_control.initial()[1]
```
Boudewijn26: I'd prefer following the line above and using destructing over indexing

Boudewijn26 (approved):

PR #421, opened 2026-07-06
Remove unneeded complexity

hmvp:
This basically solves most of the issues I had with your proposal.. 

I don't think that we should add all the complexity that is removed here just to prevent starting the connector before the listeners are registered. Besides I think we can probably solve that too..

### DIFF
--- a/zero-thrs-control/src/thrs/orchestration/comms.py
+++ b/zero-thrs-control/src/thrs/orchestration/comms.py
@@ -317,12 +317,7 @@ class ModuleMqttMapping[T: CombinedValues](MqttReceiveMapping[T]):
 
 type Publisher[T] = Callable[[T], Awaitable[None]]
 
-
-class Channels(Protocol):
-    pass
-
-
-class ControlChannels(Channels):
+class ControlChannels:
     def __init__(
         self,
         connector: "MqttConnector",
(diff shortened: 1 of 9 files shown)


zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -323,345 +318,147 @@ async def wait_for_result(self) -> T:
...
-    def _register(
-        cls,
-        connector: "MqttConnector",
-        description: D,
-    ) -> C: ...
-
-
-Ts = TypeVarTuple("Ts")
+class Channels(Protocol):
+    pass
```
Boudewijn26: I don't think this is being used

zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -670,291 +467,208 @@ class ControlApiChannels[
...
+            self._status_topic, b"", qos=1, retain=True
+        )
 
-    @staticmethod
-    def from_settings(config: "Config") -> "DirectivesApiChannelsDescription":
-        return DirectivesApiChannelsDescription(
-            controller_topic_prefix=config.mqtt_controller_topic_prefix,
+
+class DirectivesApiChannels(Channels):
+    def __init__(self, connector: "MqttConnector", config: "Config") -> None:
```
Boudewijn26, line 642: One of the things I kinda had in mind was to throw in a different connector to build up the AsyncAPI schema and then have that as an additional output. But that's a simple enough change to make when we need it: move the relevant functions to a `Protocol` and then have a `MqttConnector` and `SchemaConnector` implement those

Boudewijn26 (approved):

PR #422, opened 2026-07-08
Add THRS cosimulation and fix module bugs

aljevandam:


zero-thrs-control/src/thrs/input_output/modules/dhw.py
```diff
@@ -184,6 +184,18 @@ class DhwSensorValues(ThrsValues):
...
+        component_meta(yard_tag="50001098-01", component_type="level_switch"),
+    ]
+    dhw_level_switch_tank2: Annotated[
+        sensor.LevelSwitch,
+        component_meta(yard_tag="50001098-02", component_type="level_switch"),
+    ]
+    dhw_level_switch_tank3: Annotated[
+        sensor.LevelSwitch,
+        component_meta(yard_tag="50001098-03", component_type="level_switch"),
+    ]
```
hmvp, line 198: These already exist in this class (but with the wrong yard tag). There is only a level sensor no level switch.  Weirdly enough are the yard tags used here from the tank pressure sensors which seem to be missing...

aljevandam: All seems correct. There's level switches and level sensors, and the level sensors measure the level through pressure (so the tank pressure sensors are the tank level sensors)

zero-thrs-control/src/thrs/runtime/descriptions/simulation.py
```diff
@@ -178,6 +355,7 @@
     "high_temperature",
     "dhw",
     "thrs",
+    "thrs_control",
```
hmvp: Is this still relevant?

hmvp (approved):

zero-thrs-control/src/thrs/runtime/descriptions/simulation.py
aljevandam: no

PR #423, opened 2026-07-08
Loads symmetry

Boudewijn26:


zero-loads-app/src/loads/sensors/sail_system.py
```diff
@@ -427,7 +427,14 @@ class MainRunnerPs(LoadsModel, ABC):
...
         Load,
         Field(ge=0, le=29),
-        VariableMeta(display_name="Runner PS", scale_min=0, scale_max=29),
+        VariableMeta(
+            display_name="Runner PS",
+            scale_min=0,
+            scale_max=29,
+            side="port",
+            applies_if="windward",
+            variable_key="main-runner-load",
```
Boudewijn26: I'm only covering runners so far. This PR is just the capability, once agreed on implementation, we can extend to full modelling

zero-loads-app/src/loads/control/stub/pcan/__init__.py
Boudewijn26, line 1: Cleaning up unused code

zero-ui/src/modules/loads/graphql/queries/variables.ts
Boudewijn26, line 1: For now I'm just covering the UI enough for it not to break. Not sure how exactly we want to fit this into the design

zero-loads-app/src/loads/api/dependencies.py
```diff
@@ -84,25 +88,28 @@ async def get_reference_values(
         by_case = groupby(keys, lambda x: x[1])
         by_id_case = {}
         for case, group in by_case:
-            variables = [str(var_id) for var_id, _ in group]
+            variable_ids = [str(var_id) for var_id, _ in group]
+            variables = resolve_variable_definitions(variable_ids)
+            variable_keys = resolve_variable_keys(variables, case.wind_direction.value)
+            key_to_id = dict(zip(variable_keys, variable_ids))
             values = (
```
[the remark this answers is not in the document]
Boudewijn26: The idea is to not return reference values if they are not applicable (this is why the final `.get` defaults to `None`)

zero-ui/src/modules/loads/tests/api.integration.spec.ts
```diff
@@ -49,12 +49,14 @@ const getReferenceValues = async (
   sailset: SailId[],
   awaRange: AWA = AWA_VALUES[0].id,
   awsRange: NumRangeId = AWS_VALUES[0].id,
+  windDirection: "port" | "starboard" = "starboard",
 ) => {
```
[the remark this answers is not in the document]
Boudewijn26: That's just for testing

aljevandam (commented):
I think I understand most of this - but I do think it's a bit more omplex than necessary by introducing both 'side' and 'wind direction', and the terminology is slightly confusing. 

My suggestion would be to add 'tack' (a boat is either on the portside or starboard tack) as condition and option (instead of wind direction, which suggests TWD) and an 'applies_if' or 'applies_to_tack' to the variable meta. That should be enough info? 

So we'd have: 

Runner PT applies to the starboard tack
B Sheet SB aplies to starboard tack
etc. 

That should also simplify resolving the applicabilities which i find hard to follow. 

The tack you're on is determined by the sign of AWA (so setting it based on AWA with the existing button is easily implemented)

aljevandam (approved):
final nitpicking suggestion would be to replace applies_if with applies_to_tack (I should have been more clear before)

PR #425, opened 2026-07-08
Feature/remove combined

hmvp:
This PR removes the “combined” orchestration abstractions (e.g., CombinedModule / CombinedControl / CombinedAlarms) and updates runtime runners, MQTT channel wiring, and tests to operate on per-module ModuleDescription + per-module ControlChannels dictionaries instead.

Changes:

 -   Refactors ControlRunner/LockstepRunner to manage multiple modules via {module_name: ModuleDescription} and {module_name: ControlChannels} rather than combined wrappers.
 -   Simplifies MQTT wiring so ControlChannels is constructed for a single module (module name + ModuleDescription) and message payloads are no longer wrapped in CombinedValues for single-module use cases.
 -   Updates/removes tests and fixtures to match the new per-module APIs (including deleting the old combined-module test suite).
 -   Make sure all the SwitchingControl wrapping is centralized and knowledge about it is not spread everywhere
 -   Update SimulationTestRunner (used in tests) to use the new code so tests stay representative


zero-thrs-control/src/thrs/runtime/runners/lockstep.py
```diff
@@ -39,19 +47,20 @@ async def run(self, n_ticks: int) -> None:
...
-                sim_result.sensor_values
-            )
+            control_values_map = {}
+            for name in self.control_runner._controls:
+                control_values = await self.control_runner._execute_control_tick(
+                    name, sim_result.sensor_values.values[name]
+                )
+                if control_values:
+                    control_values_map[name] = control_values
+            self._control_values = CombinedValues(values=control_values_map)
```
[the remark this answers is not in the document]
hmvp: The first is intentional (is always programmer error). 
The second is not correct, class instances are never falsy so only None is filtered

zero-thrs-control/src/thrs/runtime/descriptions/simulation.py
```diff
@@ -190,10 +161,13 @@
         ),
         dhw_hotwater_demand=FlowBoundary(flow=Stamped.stamp(30)),
     ),
+    "adsorption": None,  # TODO Fil in  # type: ignore
```
Boudewijn26:
```suggestion
    "adsorption": None,  # TODO: Fill in  # type: ignore
```

zero-thrs-control/src/thrs/cli/cli.py
```diff
@@ -27,17 +19,22 @@
...
 
 class ControlCmd(BaseSettings):
-    mode: ModeNames
+    mode: ModeName
 
     async def cli_cmd(self) -> None:
         settings = Config()  # type: ignore
 
-        control_mode = lookup_mode(self.mode)
+        control_mode = MODES_DICT[self.mode]
```
Boudewijn26: Why switch to a dict when the function was doing just fine? (functional boundaries are preferred above value boundaries, if you insist on using a dict to speed up lookup (which might even be questionable), then you could do `lookup_mode = _MODES_DICT.get`)

zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -476,46 +472,47 @@ def __init__(
...
+        manual_values_mapping = DirectMqttMapping[C].for_module(
             module_description.control_values_cls,
             config.mqtt_controller_topic_prefix,
             module_name,
             type_topic="manual-values",
         )
         connector._register_listener(manual_values_mapping)
 
-        control_modes_mapping = DirectMqttMapping.for_module(
+        # We should not need to wrap here with switching control mode, but we need to...
```
Boudewijn26: This comment is raising more questions than it is answering. Why shouldn't we need to do it, why do we need to do it, what to do about it

zero-thrs-control/src/thrs/orchestration/comms.py
```diff
@@ -319,85 +301,99 @@ async def wait_for_result(self) -> T:
...
-        config: "Config",
-        control_module: "CombinedModule",
+        config: Config,
+        module_name: str,
+        control_module: ModuleDescription[S, C, P, M, CS],
     ) -> None:
-        sensor_values_mapping = ModuleMqttMapping(
-            control_module.sensor_values_clss,
-            PartialMqttMapping,
+        sensor_values_mapping = PartialMqttMapping[S](
```
Boudewijn26, line 318: Can't we set up the `__init__` (and `for_module` below) to derive the type from the passed argument?

zero-thrs-control/src/thrs/orchestration/module.py
```diff
@@ -45,167 +42,71 @@ def __init__(
...
-        CombinedValues,
-        CombinedValues,
-        CombinedValues,
-        CombinedValues,
-        CombinedValues,
-    ]
-):
-    """Combination of multiple controls into one control module"""
-
+class Module[
```
Boudewijn26, line 45: The functionality in here feels a bit too specific for me to have it be orchestration. I'd lean more towards putting this in runtime

zero-thrs-control/src/thrs/orchestration/module.py
```diff
@@ -45,167 +42,71 @@ def __init__(
...
+            self._control.parameters,
         )
+
+        if alarms:
+            warnings.warn(f"Alarms detected: {alarms}")  # TODO: properly handle alarms
+
+    async def send_control_updates(
+        self, sensor_values: S, control_values: C, controller_state: CS
+    ) -> None:
+        """Send control values, controller state, parameters, control modes, and manual controls to the control channels."""
```
Boudewijn26: By the time you are listing 5 items in the docstring, I think the code itself is the better documentation (also because you forgot the computed values)

zero-thrs-control/src/thrs/orchestration/setup.py
```diff
@@ -1,53 +1,74 @@
...
+    result = []
+    for module_name, module in control_modules.items():
+        parameters = module.parameters_cls()
+        control = module.control(parameters, time_fn)
 
-    parameters = {
-        module: mode.control_module.parameters_for_module(module)()
-        for module in mode.control_module.modules
-    }
+        # This line should not be here since it exposes that we are dealing with a switching module
```
Boudewijn26, line 65: Same as above, also can't you just put this into the `ModuleDescription`?

zero-thrs-control/src/thrs/orchestration/setup.py
```diff
@@ -1,53 +1,74 @@
...
+        module.control_mode_cls = SwitchingControlMode[module.control_mode_cls]
+
+        channel = ControlChannels(connector, config, module_name, module)
+
+        alarms = module.alarms()
 
-    control = mode.control_module.control(CombinedValues(parameters), time_fn)
+        result.append(Module(module_name, control, alarms, channel))
```
Boudewijn26, line 73: We generally prefer list comprehensions over stateful loops. But this might not be worth the subfunction

zero-thrs-control/src/thrs/orchestration/simulation.py
```diff
@@ -101,14 +90,62 @@ def tick(self, control_values: C) -> SimulationResult[S, C, I, O]:
...
-    def simulation_inputs(self) -> I:
-        return self._simulation_inputs
+@dataclass
+class SimulationDescription:
+    simulation_outputs_cls: type[SimulationValues]
+    fmu: Fmu
+    simulation_inputs: SimulationInputs
+
+
+class SimulationModule[
```
Boudewijn26: I'm not sure it's smart to use the term "module" for both control and simulation. Since a combination of control modules can have a single simulation module

If you are going to use it, I would suggest clearly stating that it's not necessarily one to one

zero-thrs-control/src/thrs/runtime/runners/control.py
```diff
@@ -1,83 +1,27 @@
...
-        await self._control_channels.send_control_values(control_values)
-        await self._control_channels.send_controller_state(controller_state)
-        await self._control_channels.send_parameters(self._control.parameters)
-        if self._control.mode is not None:
-            await self._control_channels.send_control_modes(self._control.mode)
-        await self._control_channels.send_manual_control(self._control.manual_controls)
+        self._modules = control_modules
+
+    async def run(self) -> None:
+        """Run control in a loop for a ticks."""
```
Boudewijn26: This isn't valid English (and I would say that it's not saying anything, of course the method `run` on the `ControlRunner` is running the control)

zero-thrs-control/src/thrs/runtime/runners/control.py
```diff
@@ -1,83 +1,27 @@
...
-        await self._control_channels.send_computed_values(sensor_values)
-        await self._control_channels.send_control_values(control_values)
-        await self._control_channels.send_controller_state(controller_state)
-        await self._control_channels.send_parameters(self._control.parameters)
-        if self._control.mode is not None:
-            await self._control_channels.send_control_modes(self._control.mode)
-        await self._control_channels.send_manual_control(self._control.manual_controls)
+        self._modules = control_modules
+
+    async def run(self) -> None:
```
Boudewijn26: If it's just a single tick, then I would say `tick` is a better name

zero-thrs-control/src/thrs/runtime/runners/lockstep.py
```diff
@@ -1,57 +1,61 @@
...
+
+            control_values, controller_state = module.execute_control_tick(
+                sensor_values
             )
+
+            await module.send_control_updates(
+                sensor_values, control_values, controller_state
+            )
+
+            control_values_map[module.name] = control_values
```
Boudewijn26: Again, we prefer comprehensions

zero-thrs-control/src/thrs/orchestration/module.py
```diff
@@ -45,167 +42,71 @@ def __init__(
...
+        return self._channels.get_sensor_values()
 
-    @property
-    def manual_controls(self) -> CombinedValues:
-        return CombinedValues(
-            values={
-                name: module.manual_controls for name, module in self._modules.items()
-            }
-        )
+    def execute_control_tick(self, sensor_values: S) -> tuple[C, CS]:
```
Boudewijn26, line 83: Executing a tick is just `tick`ing right?

Boudewijn26 (commented):
Looks good overall, just a couple of clarity items I would like to see covered

zero-thrs-control/src/thrs/orchestration/comms.py
hmvp: It already does, but that is not enough for pyright to do the right thing.. Might be a limitation of pyright...

zero-thrs-control/src/thrs/orchestration/module.py
hmvp: IMHO, the distinction between the two folders is a bit arbitrarily... Could just be one folder...  However this was already the way it was organised...

zero-thrs-control/src/thrs/runtime/runners/lockstep.py
hmvp: How would you do that here and still keep it readable?  Even if you split the body to a separate function the resulting comprehension becomes near unreadable...

zero-thrs-control/src/thrs/orchestration/module.py
hmvp: Yes but this is more readable

zero-thrs-control/src/thrs/orchestration/module.py
Boudewijn26: Most of language is arbitrary, the question is can we make it useful.

I feel like the changes you are making are pulling this away from something that is orchestrating between/within modules (what was here before) and now is something that is specifically runtime related

That isn't to say I don't agree that the difference between orchestration and runtime a clear one. I wouldn't be opposed to pulling stuff out of orchestration into specifics and putting the rest into runtime (not a biggest fan of the term orchestration)

zero-thrs-control/src/thrs/orchestration/simulation.py
hmvp: That already follows from how and where you can use it (it does not share an interface)

zero-thrs-control/src/thrs/runtime/runners/lockstep.py
Boudewijn26:
```python
async def tick(self) -> None:
        """Run simulation and control in lockstep for a tick.
        Retrieve parameters and automation modes from the control channels, and simulation inputs from the simulation channels.
        Send control values to the simulation channels, and sensor values to the control channels.
        """
        # We are ignoring the sensor values here since we get them from the simulation result
        for module in self.control_modules:
            await module.sync_control_channels_state()

        self.simulation_module.sync_simulation_inputs()

        sim_result = self.simulation_module.execute_simulation_tick(
            self._control_values
        )

        await self.simulation_module.send_simulation_updates(sim_result)

        control_values_map = {
            module.name: control_values
            for module in self.control_modules
            if (control_values := await self._tick_module(module, sim_result))
        }

        self._control_values = CombinedValues(values=control_values_map)

    async def _tick_module(
        self, module: Module, sim_result: SimulationResult[S, CombinedValues, I, O]
    ) -> ThrsValues | None:
        sensor_values = sim_result.sensor_values.values.get(module.name)

        if sensor_values is None:
            return None

        control_values, controller_state = module.execute_control_tick(sensor_values)

        await module.send_control_updates(
            sensor_values, control_values, controller_state
        )

        return control_values
```

zero-thrs-control/src/thrs/orchestration/simulation.py
Boudewijn26: Still naming it `SimulationModule` implies some connection to the `Module`. Sure, you can deduce that they are different, but then you have to do the work while reading the code and then backtrack; it violates POLA. Can't we name it `SimulationUnit` or something similar to prevent that confusion?

zero-thrs-control/src/thrs/runtime/runners/lockstep.py
hmvp: Yeah I did that too, but I am not convinced that this is more readable...  Also you are missing an extra `is not None` which makes it even worse...

zero-thrs-control/src/thrs/orchestration/simulation.py
hmvp: There is a connection, they are both very similar in API and contain logic + channels. But SimulationUnit is fine with me

zero-thrs-control/src/thrs/orchestration/module.py
hmvp: I do agree, but it adds quite a lot of extra changes to an already big PR, so I would like to reserve that for a future refactor...

zero-thrs-control/src/thrs/runtime/runners/lockstep.py
Boudewijn26: Which `is not None` am I missing?

This takes your > 30 LOC function and cleanly splits it up. And it is codebase convention. I see a difference between local readability and linear readability.

zero-thrs-control/src/thrs/orchestration/module.py
Boudewijn26: Yeah, I wasn't arguing to put it in here

zero-thrs-control/src/thrs/runtime/runners/lockstep.py
hmvp: in ` if (control_values := await self._tick_module(module, sim_result))` the result can be None, but we don't want that in the CombinedValues instance..

I am for splitting the function off, and generally I prefer comprehensions, but not if the become multi line monsters that use every trick available... I'd rather keep the for loop in that case to keep it readable

hmvp: Ah technically the `if` already checks that... I am so used to make my None checks explicit, but instances can never be Falsy...

hmvp: Changed

Boudewijn26: If the comprehension becomes too complicated, I generally try to split the comprehension into separate variables (dicts) which are then merged

zero-thrs-control/src/thrs/orchestration/setup.py
hmvp: Nope, because this is needed for the channels which already exist by the time the module init is called..

zero-thrs-control/src/thrs/orchestration/setup.py
hmvp: While I do agree in the other case where there is a result to fill. In this case I would never use a comprehension. Comprehesions signal that you have a result (a list/dict etc) and we don't here

zero-thrs-control/src/thrs/orchestration/setup.py
hmvp: Amended

zero-thrs-control/src/thrs/orchestration/setup.py
Boudewijn26: Uhm, you are literally calling `append` on a `list` called `result`

hmvp: Ugh, you are correct...  Something something multitasking... But yeah this not worth the extra function now.. I promise to revisit this if I find a nice way to handle the switching control...

Boudewijn26 (approved):

PR #426, opened 2026-07-09
Chore/cleanups

hmvp:


Boudewijn26 (approved):
Regression test on the setting of manual values would be nice

PR #433, opened 2026-07-13
Make loads app not crash on validation errors in incoming data

Boudewijn26:


zero-loads-app/src/loads/api/messaging.py
```diff
@@ -70,17 +71,26 @@ async def _run(self):
...
 
     def _match_receiver(self, message: Message) -> MessageReceiver | None:
         return self._receivers.get(message.topic.value, None)
 
-    def _parse_message[T: LoadsModel](self, message: Message, model: type[T]) -> T:
+    def _parse_message[T: LoadsModel](
+        self, message: Message, model: type[T]
+    ) -> T | None:
         if not isinstance(message.payload, str | bytes):
             raise ValueError(f"Expected string or bytes, got {type(message.payload)}")
```
hmvp, line 86: This still breaks your listener, with proper error handing in place you might also want to return None here too

Boudewijn26: Empty strings aren't expected, so I say we cross that bridge when we get to it

hmvp (changes requested):

hmvp (approved):

PR #434, opened 2026-07-13
ZERO-2038: bootstrap thrusters mimic

null-33:


zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -162,6 +162,17 @@ def thrusters_temperature_recovery(self) -> sensor.CalculatedTemperature:
             )
         )
 
+    @computed_field(
+        json_schema_extra=computed_meta(
+            yard_tag="50001001", component_type="heat_exchanger", included_in_fmu=False
+        )
+    )
+    @property
+    def thrusters_seawater_exchanger(self) -> sensor.HeatExchanger:
```
aljevandam, line 169: If I recall correcty, the delta t for the exchanger itself is the delta from in and out in THIS circuit. So the return temp is 1038-28, and the supply temp is either 1038-30 OR a weighted average of 1038-01 and 1038-02, depending on the switch valve positions. 

The flow also needs to computed I guess (1218-2 + 1057-22 - 1218-01?)

Looks like it will be quite messy but its the best we can do I think..



A pattern similar to 
```
@property
    def dhw_consumers_exchanger(self) -> sensor.HeatExchanger:
        if sensor.valves_open_closed(
            open_valves=[self.dhw_switch_high_temperature],
            closed_valves=[
                self.dhw_switch_heatpump,
                self.dhw_switch_low_temperature,
            ],
        ):
            return sensor.HeatExchanger.from_sensors(
                temperature_supply=self.dhw_temperature_boosting_supply.temperature,
                temperature_return=self.dhw_temperature_boosting_return.temperature,
                flow=self.dhw_flow_boosting.flow,
                heat_transfer_conversion=WATER_HEAT_TRANSFER_CONVERSION,
            )
        else:
            return sensor.HeatExchanger(delta_t=Stamped.stamp(0), heat=Stamped.stamp(0))
```

The values for the cicruit box of the seawater requires seawater temperature, return temperature and seawater flow (but we might decide to just show seawater temperature there if that data is limited..)

aljevandam (commented):
I reviewed the python part only

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
hmvp: Yeah we also might change the design to reflect that we don't really know these values.. Becuase the frontend components require a exchanger to be present I added this dummy one, but when we find a good solution it might be removed completely

aljevandam: I don't think we are aligned here. The seawater exchanger values we definitely want to see (dt and kW), and we know the values (we just need to compute them as above) 

For the temperatures and flows on the seawater side (the "SEAWATER" box) we might need to shift to showing what we know. 

image

zero-thrs-control/src/thrs/control/modules/thrusters.py
```diff
@@ -1,18 +1,20 @@
...
-from typing import Literal, NoReturn
+from typing import Annotated, Literal, NoReturn
 
 from pydantic import model_validator
 from transitions import Machine, State
 
 from thrs.classes.control import Control, ControlMode
 from thrs.classes.machine_state_logger import StateLogger
 from thrs.control.controllers import FlowBalanceController, PidController
+from thrs.control.modules.dhw import _zero_pid
```
Boudewijn26: `_` indicates private (also thrusters shouldn't depend on dhw)

zero-ui/scripts/README.md
```diff
@@ -32,24 +32,26 @@ pnpm extract-schema-values <CONST_NAME> <TYPE_NAME>
 
 **What it does:**
 
-1. Parses the GraphQL schema file at `src/modules/thrs/graphql/schema.graphql` using the official GraphQL parser
+1. Parses the GraphQL schema file at `src/modules/thrsim/graphql/schema.graphql` using the official GraphQL parser
```
Boudewijn26, line 35: The schema isn't specific to `thrsim`

zero-ui/src/modules/thrapp/mimics/modules/thrusters/data/exchange-circuits/_pcm.ts
```diff
@@ -0,0 +1,25 @@
...
+  sensors: {
+    flowIn: getField(SensorComponentType.Flow, "thrusters", "thrustersFlowRecovery"),
+    flowOut: getField(SensorComponentType.Flow, "thrusters", "thrustersFlowRecovery"),
+    tIn: getField(SensorComponentType.Temperature, "thrusters", "thrustersTemperatureRecoveryMix"),
+    tOut: getField(SensorComponentType.Temperature, "thrusters", "thrustersTemperatureSupply"),
+  },
+  get tooltip() {
+    return fieldTooltip(this.source, {
+      title: "PCM",
+      componentType: "Exchange circuit",
```
Boudewijn26, line 22: These look user facing, so technically should be i18ned, also below

Boudewijn26 (approved):

zero-ui/scripts/README.md
hmvp: Nope, but it does live there currently

zero-ui/src/modules/thrapp/mimics/modules/thrusters/data/exchange-circuits/_pcm.ts
hmvp: I agree, but there are a couple more places where we need to fix that.. So I opted for consistency over correctness here. Also under the assumption that this needs to be fixed for all places in one go

PR #440, opened 2026-07-14
Fix usb hotplugging for kiosk on (sub)zero

hmvp:


Boudewijn26 (approved):
Let's see how well this works

PR #442, opened 2026-07-15
Fix control not sending control values if sensor values are None

hmvp:
And some other cleanups

This prevents control and simulation from running when not started in lockstep


zero-thrs-control/tests/orchestration/test_module.py
```diff
@@ -0,0 +1,81 @@
+from unittest import mock
+
+from thrs.classes.control import Control
+from thrs.control.switching import AutomationMode
+from thrs.input_output.alarms import BaseAlarms
+from thrs.orchestration.module import Module
+
+
+async def test_module_sends_control_when_sensor_values_are_none():
```
Boudewijn26: This only checks the return value, not what was sent

zero-thrs-control/tests/orchestration/test_module.py
```diff
@@ -0,0 +1,81 @@
...
+    )
+
+    control_values = await module.tick(None)
+
+    assert control_values == mock.sentinel.control_values
+    assert mock_control.control.call_count == 0
+    assert mock_alarms.check.call_count == 0
+
+
+async def test_module_sends_initial_control_when_manual():
```
Boudewijn26: Same

zero-thrs-control/tests/orchestration/test_module.py
```diff
@@ -0,0 +1,81 @@
...
+    control_values = await module.tick(mock.sentinel.sensor_values)
+
+    assert control_values == mock.sentinel.control_values
+    assert mock_control.control.call_count == 0
+    assert mock_alarms.check.call_args_list == [
+        mock.call(mock.sentinel.sensor_values, mock.sentinel.control_values, mock.ANY)
+    ]
+
+
+async def test_module_sends_control_when_automatic():
```
Boudewijn26: Same, though this does do other checks

Boudewijn26 (approved):
LGTM apart from the tests not doing what they say they do

zero-thrs-control/tests/orchestration/test_module.py
hmvp: It does actually, the mock.sentinel value must match the value as set in the mock initial call

PR #447, opened 2026-07-16
Add liveness check to thrs deploy

hmvp:


zero-thrs-control/charts/zero-thrs-control/templates/simulation-deployment.yaml
```diff
@@ -53,6 +55,14 @@ spec:
...
             - mountPath: /tmp
               name: tmp-volume
+          livenessProbe:
+            exec:
+              command:
+              - /bin/bash
+              - -c
+              - "test \"$(( `date +%s` - `stat -L --format %Y {{ .Values.livenessPath }}` ))\" -lt 10"
+            initialDelaySeconds: 30
+            periodSeconds: 5
```
Boudewijn26: Since this is duplicated with the control-deployment, you could consider introducing a helper for this

zero-thrs-control/charts/zero-thrs-control/values.yaml
```diff
@@ -12,6 +12,8 @@ mqtt:
   controllerTopicSuffix: set
   simulatorTopicSuffix: set
 
+livenessPath: /tmp/liveness
```
Boudewijn26, line 15: Wouldn't it make more sense to put the liveness path under both control and simulation? It doesn't apply to API

zero-thrs-control/src/thrs/runtime/runners/control.py
```diff
@@ -9,11 +10,16 @@ def __init__(
...
             Module[S, ThrsValues, ThrsValues, ThrsValues, ThrsValues]
         ],
+        liveness: Liveness,
     ):
         self._modules = control_modules
+        self._liveness = liveness
 
     async def tick(self) -> None:
         """Run control for a tick."""
+        if self._liveness:
```
Boudewijn26: `self._liveness` is always thruthy, so this doesn't do anything (probably a remnant from an earlier time)

zero-thrs-control/tests/orchestration/test_liveness.py
```diff
@@ -0,0 +1,23 @@
+import os
+from tempfile import NamedTemporaryFile
+
+from thrs.runtime.liveness import Liveness
+
+
+def test_liveness_no_error():
+    liveness = Liveness("/tmp/does_not_exist")
+
+    # Does not raise
```
Boudewijn26: [The pytest docs propose using `nullcontext` to make a does-not-raise explicit](https://docs.pytest.org/en/stable/example/parametrize.html#parametrizing-conditional-raising)

```suggestion
    from contextmanager import nullcontext as assert_does_not_raise
    
    with assert_does_not_raise():
      liveness.signal()
```

Boudewijn26 (commented):
Just enough nitpicks that I can't give this the all clear

zero-thrs-control/charts/zero-thrs-control/values.yaml
hmvp: No we use the same ConfigMap for everything so it needs to be "global"

Boudewijn26 (approved):

zero-thrs-control/tests/orchestration/test_liveness.py
```diff
@@ -0,0 +1,24 @@
...
+from contextlib import nullcontext as assert_does_not_raise
+from tempfile import NamedTemporaryFile
+
+from thrs.runtime.liveness import Liveness
+
+
+def test_liveness_no_error():
+    liveness = Liveness("/tmp/does_not_exist")
+
+    with assert_does_not_raise:
```
Boudewijn26:
```suggestion
    with assert_does_not_raise():
```

Boudewijn26 (commented):
This'll fix your test

PR #450, opened 2026-07-17
Add mapping to sailpack variables and seed export

Boudewijn26:
Also introduces proper test data provisioning so the seed can change without breaking the tests


zero-loads-app/src/loads/api/model.py
```diff
@@ -84,6 +74,113 @@ async def get_loads_reference_values(
...
+        grouped[str(load_case_id)].append(
+            ReferenceValue(
+                id=ref_value.variable_key,  # type: ignore
+                alarm_low=ref_value.alarm_low,  # type: ignore
+                warning_low=ref_value.warning_low,  # type: ignore
+                target=ref_value.target,  # type: ignore
+                warning_high=ref_value.warning_high,  # type: ignore
+                alarm_high=ref_value.alarm_high,  # type: ignore
+            )
+        )
```
Boudewijn26, line 144: Unfortunately this `defaultdict` pattern doesn't translate well to comprehensions (`groupby` is finicky)

hasura/migrations/zero/20260424000200_loads_schema/up.sql
florissmit10, line 1: Arent you supposed to make a new migration for this kind of stuff?
Or do we start doing that when we go "live"?

Boudewijn26: Technically yes, but since we're not live yet this is OK (and keeps the schema clearer)

zero-loads-app/tests/conftest.py
```diff
@@ -32,8 +32,8 @@ def pytest_runtest_setup(item):
 def settings():
     from loads.config import Settings
 
-    os.environ["PG_HOST"] = "localhost"
-    os.environ["MQTT_HOST"] = "localhost"
+    os.environ.setdefault("PG_HOST", "localhost")
```
florissmit10, line 35: Doesn't this pollute the env variables?

florissmit10 (approved):

PR #461, opened 2026-07-27
Get UI mimic components information from api definitions

hmvp:


zero-ui/src/modules/thrapp/components/legends/items/HeatPump.vue
```diff
@@ -18,10 +18,7 @@ const assets = DHW_MIMIC_DATA[MimicComponentType.HeatPump];
 <template>
   <LegendItem class="min-h-27">
     <LegendItemPreview>
-      <HeatPumpInstance
-        tag-id="1035"
-        v-bind="assets['1035']"
-      />
+      <HeatPumpInstance v-bind="assets['1035']" />
```
Boudewijn26, line 21: Does the heat pump not have a technical name we can index it by?

zero-ui/src/modules/thrapp/components/tooltip/index.ts
```diff
@@ -73,7 +74,9 @@ export const createTooltipContext = (
     for (const [type, data] of typesData) {
       const values = Object.values(data) as TooltipComponentContext[];
 
-      const sourceValue = values.find((val) => isEqual(val.source, source));
+      const sourceValue = values.find(
+        (val) => val.source && isEqual(val.source.slice(0, 3), source.slice(0, 3)),
```
Boudewijn26: Why only compare the first 3?

zero-ui/src/modules/thrapp/mimics/modules/dhw/layers/ActuatedValves.vue
```diff
@@ -21,7 +21,7 @@ const flowControlValves = DHW_MIMIC_DATA[MimicComponentType.FlowControlValve];
       <TagLabelInstance
         x="360"
         y="192"
-        tag-id="1067-11"
+        :tag-id="switchValves['1067-11'].tooltip?.yardTag"
```
Boudewijn26, line 24: This feels a bit circular. Can't we index by technical name instead?

zero-ui/src/modules/thrapp/mimics/providers/index.ts
```diff
@@ -244,7 +251,7 @@ export const getDefinition = (field: ModuleField) => {
   } else if (isControllerStateField(field)) {
     return getControllerStateDefinition(field[1], field[2]);
   } else if (isCustomField(field)) {
-    return undefined;
+    return field[3];
```
Boudewijn26: Prefer destructuring over magic indexing (though not a blocker, because that's already happening above)

zero-ui/src/modules/thrapp/components/tooltip/index.ts
hmvp: the fourth is a new optional item that contains extra data for custom fields (basically fields that are not present in the api like manual valves or boiler tanks)

zero-ui/src/modules/thrapp/mimics/modules/dhw/layers/ActuatedValves.vue
hmvp: Currently everything is indexed using the yard tag.. We might want to change that at some point but that is way outside the scope of this change

zero-ui/src/modules/thrapp/components/legends/items/HeatPump.vue
hmvp: The heatpump does, but this is how the whole frontend refers to these components atm. Som components don't though, they only have a yard tag (like manual valves)

zero-ui/src/modules/thrapp/components/tooltip/index.ts
Boudewijn26: Would it be worthwhile to destruct to make that a bit clearer?

hmvp: No not really due to the way we handle it here. But there is another improvement that (re)moves the slice to the central place that deals with  how this is structured. Changed

Boudewijn26 (approved):

PR #463, opened 2026-07-27
Update greptime & add GOOGLE_APPLICATION_CREDENTIALS env variable

florissmit10:
ZERO-1653

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -169,9 +169,12 @@ services:
         condition: service_healthy
 
   greptimedb:
-    image: greptime/greptimedb:v1.0.0
+    image: greptime/greptimedb:v1.1.3
     security_opt:
       - seccomp=unconfined
+    environment:
+      # You can copy in credentials for importing data from gcp
+      GOOGLE_APPLICATION_CREDENTIALS: /gcp-credentials.json
     ports:
       - "4000-4003:4000-4003"
     restart: always
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -173,7 +173,8 @@ services:
     security_opt:
       - seccomp=unconfined
     environment:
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -170,9 +169,12 @@ services:
         condition: service_healthy
 
   greptimedb:
-    image: greptime/greptimedb:v1.0.0
+    image: greptime/greptimedb:v1.1.3
     security_opt:
       - seccomp=unconfined
+    environment:
+      # You can copy in credentials for importing data from gcp
+      GOOGLE_APPLICATION_CREDENTIALS: /gcp-credentials.json
```
hmvp, line 178: How does this work? is that file already in the image?

florissmit10: There is no file in the image and without a file and an import command it does not really do anything. 
The idea is that you can copy the one in that you need (zero, singel, subzero). See https://github.com/foundation-zero/edge-flux/pull/257

I tried adding this env variable in the script, but it needs the main process to have it, instead of only the shell where you do the import.

Boudewijn26: I think the comment above needs to be clearer that it requires the user to manually copy the file into the docker container. As it stands it reads more like a feature of the image

Boudewijn26 (commented):
See comment in thread

docker-compose.yml
hmvp: Wouldn't it be better to use a mount then?

florissmit10: Most people wont use the functionality, so my thought was that its better to be optional. Otherwise people will try to start greptime, and suddenly they have to download a gcp key for that to work.

Most of the logic of this stuff is in the edge-flux backup readme, but this was something which cant be done from there.

hmvp (approved):

PR #466, opened 2026-07-28
Add shared modbus bridge and power-tags ingestion

Boudewijn26:
- New zero-modbus-bridge package: FastStream-based Modbus TCP to MQTT bridge with Float32/uint16/bool support, invalid-value handling, per-unit host overrides, and AsyncAPI 3.0.0 generation via MQTTBroker publisher registration
- New zero-power-tags: parses 5 .xlsm files (226 breakers across 5 gateways) into modbus_bridges.json with human-readable topic names; supports patches.json for consumer-name corrections
- Refactor zero-termodinamica and zero-hull-temperature to use the shared bridge; delete duplicated bit_ops/settings/stub/bridge
- Add power-tags services to docker-compose under data-collection profile, plus MQTT topics in vector config
- 24 typo corrections in power tag consumer names via patches.json (BATERY, DRAINK, PANLES, DISTRBUTION, etc.)


.github/workflows/ci_modbus_bridge.yaml
```diff
@@ -0,0 +1,33 @@
+name: "Modbus Bridge: Test"
+on:
+  pull_request:
+    paths:
+      - zero-modbus-bridge/**
```
florissmit10, line 5: This path should also be added to the other two gha files right? IF the library code changes you want to trigger their CI. Or does UV handle that somehow?

zero-hull-temperature/tests/test_bit_ops.py
```diff
@@ -1,4 +1,4 @@
-from zero_hull_temperature.bit_ops import float_to_lsw_registers, lsw_registers_to_float
+from zero_modbus_bridge.bit_ops import float_to_lsw_registers, lsw_registers_to_float
```
florissmit10: Shouldnt the tests move if the functions move?

zero-modbus-bridge/src/zero_modbus_bridge/settings.py
```diff
@@ -1,11 +1,8 @@
-from contextlib import asynccontextmanager
-from typing import AsyncGenerator
+from typing import Any
 
-from aiomqtt import Client as MqttClient
-from pydantic_settings import (
-    BaseSettings,
-    SettingsConfigDict,
-)
+from faststream.mqtt import MQTTBroker
```
florissmit10, line 3: New broker client? How do you like it so far?

zero-modbus-bridge/src/zero_modbus_bridge/settings.py
```diff
@@ -26,15 +23,11 @@ class MqttSettings(BaseSettings):
...
-    @asynccontextmanager
-    async def mqtt_client(self) -> AsyncGenerator[MqttClient, None]:
-        async with MqttClient(
-            self.mqtt_host,
-            self.mqtt_port,
-            username=self.mqtt_username,
-            password=self.mqtt_password,
-        ) as mqtt:
-            yield mqtt
+    def make_broker(self) -> MQTTBroker:
```
florissmit10, line 26: I see we move from async to sync. Makes sense given the cyclic nature of modbus. Do you also envision to use this library in other async apps?

zero-modbus-bridge/src/zero_modbus_bridge/reader.py
```diff
@@ -0,0 +1,121 @@
...
+from zero_modbus_bridge.io import ModbusField, ModbusTopic, apply_modbus_field
+
+logger = logging.getLogger(__name__)
+
+# Invalid Float32 sentinel used by Schneider PowerTag and Moore HES controllers.
+# 0xFFC00000 → registers [0xFFC0, 0x0000]
+INVALID_FLOAT_SENTINEL = 0xFFC00000
+
+
+def _is_invalid_float(regs: list[int]) -> bool:
```
florissmit10: maybe add this to the bit_ops.py (and rename it)

zero-modbus-bridge/src/zero_modbus_bridge/io.py
```diff
@@ -0,0 +1,135 @@
...
+    (relative to ``ModbusTopic.start_register``), never both.
+    """
+
+    register: int | None = None
+    offset: int | None = None
+    count: int = 1
+    data_type: Literal["uint16", "float32", "bool"] = "uint16"
+    modbus_type: Literal["holding", "coil"] = "holding"
+    scale_factor: float = 1.0
+    invalid_value: int | None = None
```
florissmit10: What does this do? It filters out a single invalid value that becomes null? Shouldnt that be typed to the dataType.
Wouldnt it be better/more flexbile to pass in a validation function?

zero-modbus-bridge/src/zero_modbus_bridge/io.py
```diff
@@ -0,0 +1,135 @@
...
+        self.data_type = data_type
+        self.modbus_type = modbus_type
+        self.scale_factor = scale_factor
+        self.invalid_value = invalid_value
+
+
+class ModbusTopic[T: BaseModel](BaseModel):
+    """Describes one MQTT topic backed by Modbus registers.
+
+    Two modes:
```
florissmit10: Wouldnt it be better to introduce two sub classes for this?

zero-modbus-bridge/poetry.lock
florissmit10, line 1: Leftover from poetry to uv migration?

zero-modbus-bridge/src/zero_modbus_bridge/publisher.py
```diff
@@ -0,0 +1,45 @@
...
+    def register_publishers(
+        broker: MQTTBroker,
+        topics: list[ModbusTopic],
+    ) -> dict[str, MQTTPublisher]:
+        """Register one FastStream publisher per topic, return ``topic→publisher`` dict."""
+        publishers: dict[str, MQTTPublisher] = {}
+        for topic in topics:
+            publisher = broker.publisher(
+                topic.topic,
+                schema=topic.model,
```
florissmit10, line 34: Does this handle schema validation?

zero-modbus-bridge/src/zero_modbus_bridge/reader.py
```diff
@@ -0,0 +1,121 @@
...
+    Topics are expected to provide both ``fields`` and ``converter``.
+    Annotation-driven topics are pre-compiled into those two values by
+    ``ModbusTopic.model_post_init``.
+    """
+
+    def __init__(self, modbus: ModbusClient, topics: list[ModbusTopic]):
+        self._modbus = modbus
+        self._topics = topics
+
+    async def read_all(self) -> AsyncIterator[tuple[str, Any]]:
```
florissmit10: Why is this async?

zero-hull-temperature/Dockerfile
```diff
@@ -3,36 +3,33 @@ ENV UV_COMPILE_BYTECODE=1 UV_LINK_MODE=copy
 ENV UV_NO_DEV=1
 ENV UV_PYTHON_DOWNLOADS=0
 
+RUN apt-get update && apt-get install -y --no-install-recommends git && rm -rf /var/lib/apt/lists/*
```
florissmit10: Why does it need git?

zero-hull-temperature/zero_hull_temperature/stub.py
florissmit10, line 1: Shouldn't this use the shared stub?

zero-power-tags/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+        {{- include "zero-power-tags.labels" . | nindent 8 }}
+        {{- with .Values.podLabels }}
+        {{- toYaml . | nindent 8 }}
+        {{- end }}
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      securityContext:
```
florissmit10, line 29: I dont think all of these are required, avoid respecifying default values

zero-power-tags/charts/templates/deployment.yaml
```diff
@@ -0,0 +1,93 @@
...
+        fsGroupChangePolicy: Always
+        sysctls: []
+        supplementalGroups: []
+        fsGroup: 1001
+      containers:
+        - name: power-tags
+          args: ["run"]
+          securityContext:
```
florissmit10, line 34: I dont think all of these are required, avoid respecifying default values

zero-power-tags/charts/values.yaml
```diff
@@ -0,0 +1,34 @@
+replicaCount: 1
+
+image:
+  repository: europe-west1-docker.pkg.dev/common-449414/common/zero-power-tags
+  pullPolicy: IfNotPresent
+  tag: ""
+
+mqtt:
+  host: ""
```
florissmit10, line 9: i'd say the default can be the rabbitmq k8s url since its the same on all envs

zero-power-tags/README.md
```diff
@@ -0,0 +1 @@
+# zero-power-tags
```
florissmit10: Content missing

zero-modbus-bridge/README.md
```diff
@@ -0,0 +1 @@
+# zero-modbus-bridge
```
florissmit10, line 1: Content missing

zero-termodinamica/Dockerfile
```diff
@@ -3,29 +3,26 @@ ENV UV_COMPILE_BYTECODE=1 UV_LINK_MODE=copy
 ENV UV_NO_DEV=1
 ENV UV_PYTHON_DOWNLOADS=0
 
+RUN apt-get update && apt-get install -y --no-install-recommends git && rm -rf /var/lib/apt/lists/*
```
florissmit10: Why is git added?

florissmit10: Ah for the version tag by any chance?

zero-modbus-bridge/src/zero_modbus_bridge/settings.py
Boudewijn26: It's fine enough. Python doesn't really have a great pattern to document willy nilly `client.publish`s. This one comes closest with its publisher objects. Mostly I want asyncapi docs for everything so we can generate APIs from it and clean up our data management

zero-modbus-bridge/src/zero_modbus_bridge/settings.py
Boudewijn26: If by this library you mean Zero Modbus Bridge, only for the modbuses. If you mean faststream, yes, maybe. I want every application to output asyncapi to describe its MQTT behavior. FastStream can help, but might not fit

Also this is just the description of the broker (think fastapi `Router`), the app itself is still async

zero-power-tags/charts/values.yaml
Boudewijn26: I don't think zero should be that aware of our flux config (unless we want to merge the zero flux into this repo)

zero-modbus-bridge/src/zero_modbus_bridge/publisher.py
Boudewijn26: No, unfortunately not. Might be something we can fix for them

zero-hull-temperature/Dockerfile
Boudewijn26: It doesn't

florissmit10 (approved):

PR #468, opened 2026-07-28
Migrate all Python projects from Poetry to uv

Boudewijn26:
Migrate all 7 Python services from Poetry to uv.

- Run uvx migrate-to-uv on all 7 Python projects
- Switch build backend to hatchling+hatch-vcs
- Generate uv.lock files, remove poetry.lock files
- Rewrite 6 Dockerfiles to use astral-sh/uv builder images
- Rewrite 7 CI workflows (setup-uv, uv sync, uv run, uvx hatch version)
- Update justfiles: poetry run -> uv run
- Update READMEs, CODESTYLE.md, .dockerignore, .vscode settings
- Update thrs-control README to use just commands


Boudewijn26:
PEP 508 note: the Poetry-style parenthesized version specifiers (e.g. `aiomqtt (>=2.3.0,<3.0.0)`) are handled correctly by uv, pip, and hatchling — they're valid PEP 440 constraints even if the parens are non-canonical. The migration tool preserved them as-is and they resolve correctly. This is a cosmetic preference, not a functional issue.

zero-data/Dockerfile
```diff
@@ -1,28 +1,18 @@
...
+ENV UV_NO_DEV=1
+ENV UV_PYTHON_DOWNLOADS=0
 
 WORKDIR /app
-COPY pyproject.toml poetry.lock README.md ./
-RUN poetry install --no-root
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --locked --no-install-project
```
hmvp: you probably want to add `--no-editable` 

Same for all other docker files..

You might also want to build/install the app itself in the builder image, but you need to make sure that packaging is correct (see https://docs.astral.sh/uv/guides/integration/docker/#non-editable-installs) 
Otherwise you are going to mis non python files

zero-thrs-control/README.md
```diff
@@ -11,57 +11,54 @@ cp .env.example .env
...
+brew install graphviz
 ```
 
 ## Lockstep
 
 Run the lockstep with:
 
 ```bash
-poetry run python -m thrs.cli lockstep --mode <module>
+just run_lockstep module=dhw
```
hmvp: `module=` is not needed (not even sure it works)

same for the rest of the file

hmvp (approved):

zero-data/pyproject.toml
```diff
@@ -71,12 +67,13 @@ select = ["E4", "E7", "E9", "F", "W", "I"]
 [tool.dotenv]
 location = ".env"
 
-[tool.poetry.requires-plugins]
-poetry-dynamic-versioning = { version = ">=1.0.0,<2.0.0", extras = ["plugin"] }
+[tool.hatch.version]
+source = "vcs"
```
florissmit10, line 71: Does this replace the dynamic versioning behaviour?

Boudewijn26: Yes, that needs some wrangling to keep `uv` happy

zero-domestic-control/src/domestic_control/services/hass.py
```diff
@@ -243,8 +243,16 @@ def __init__(self, hass: HassWsClient, data: Sink):
     def _run(self, loop: asyncio.AbstractEventLoop):
         with self._hass.listen_events("state_changed") as events:
             for event in events:
-                if event.data["entity_id"].startswith("input_number."):
-                    number_change = InputNumberChanged(**event.data)
+                event_data = event.data
```
florissmit10, line 246: Why where these python files touched? Seems out of scope

Boudewijn26: Type checking wasn't happy. We upgraded pyright

florissmit10 (approved):

PR #472, opened 2026-07-30
Add load case mapping and show selected load case in frontend

aljevandam:


zero-loads-app/src/sailpack/export_load_case_mapping_seed.py
```diff
@@ -0,0 +1,157 @@
...
+                    ),
+                )
+                records.append(
+                    {
+                        "load_case_id": best["id"],
+                        "awa_range_id": awa_range_id,
+                        "aws_range_id": aws_range_id,
+                        "sail_abbreviations": list(sail_abbreviations),
+                    }
+                )
```
Boudewijn26: We do still prefer comprehensions over mutable iteration (especially when the body is pretty big)

zero-loads-app/src/sailpack/export_load_case_mapping_seed.py
```diff
@@ -0,0 +1,157 @@
...
+
+AWS_RANGES: dict[int, tuple[float, float | None]] = {
+    1: (0.0, 10.0),
+    2: (10.0, 15.0),
+    3: (15.0, 20.0),
+    4: (20.0, 25.0),
+    5: (25.0, 30.0),
+    6: (30.0, 40.0),
+    7: (40.0, None),
+}
```
Boudewijn26: Why not just `list` and `enumerate` (and accept 0-indexing)?

zero-loads-app/src/sailpack/export_load_case_mapping_seed.py
```diff
@@ -0,0 +1,157 @@
...
+                "tws": tws,
+                "twa": twa,
+                "aws": aws,
+                "awa": awa,
+                "bsp": bsp,
+                "heel": heel,
+                "sail_abbreviations": sorted(abbreviations),
+            }
+        )
+    return cases
```
Boudewijn26: Also this can be a comprehension

zero-loads-app/src/sailpack/export_reference_values_seed.py
```diff
@@ -0,0 +1,210 @@
...
+                "name": calculation_id,
+                "tws": tws,
+                "twa": twa,
+                "aws": aws,
+                "awa": awa,
+                "bsp": bsp,
+                "heel": heel,
+                "sail_abbreviations": sail_abbreviations,
+            }
+        )
```
Boudewijn26, line 49: Can be a comprehension (that would also move `load_case_records` to the more sensible place)

zero-ui/src/modules/loads/stores/variables.ts
```diff
@@ -216,12 +221,22 @@ export const useVariablesStore = defineStore("loads-variables", () => {
...
   });
 
+  const selectedLoadCase = computed(() => referenceValues.value?.loadCase ?? null);
+  const selectedLoadCaseLabel = computed(() => {
+    if (!selectedLoadCase.value) {
+      return null;
+    }
+
+    const { name, awa, aws, heel, twa, tws } = selectedLoadCase.value;
+    return `Sailpack model: ${name} AWA: ${formatAwa(awa)}° AWS: ${formatAws(aws)} kts Heel: ${formatHeel(heel)}° TWA: ${formatTwa(twa)}° TWS: ${formatTws(tws)} kts`;
```
Boudewijn26: This is perhaps more view/component logic than something that belongs in a store

zero-ui/src/modules/loads/stores/variables.ts
```diff
@@ -60,6 +60,11 @@ const toSupportedValue =
     supportedValues.includes(storedValue) ? storedValue : defaultValue;
 
 const extractId = extractProperty("id");
+const formatAwa = (awa: number) => awa.toFixed(1);
+const formatAws = (aws: number) => aws.toFixed(1);
+const formatHeel = (heel: number) => heel.toFixed(1);
+const formatTwa = (twa: number) => twa.toFixed(1);
+const formatTws = (tws: number) => tws.toFixed(1);
```
Boudewijn26: Technically we should i18n this, but if you move it to the view as suggested, it's probably fine

PR #477, opened 2026-08-03
Remove dataframe related code from "production code" and keep it in notebook related code

hmvp:
The dataframe functionality made some stuff not work properly and was not very nice for the type checkers..

The new version does the same without needing all the code to adapt to it


aljevandam (approved):
Look good. 

This does prevent us from using de dataframes in thrsim at some point but perhaps that should use a different approach either way. 

I think remocving the messy parts from produciton code is a good idea.

hmvp:
> This does prevent us from using de dataframes in thrsim at some point but perhaps that should use a different approach either way.
> 
> I think remocving the messy parts from produciton code is a good idea.

Yeah the trouble with that is that dataframes are not really suited to be passed through graphql and if we need to design a different way to transfer them then we can connect that to the actual simulation in a similar way as in this PR (eg. pass though graphql some generator definition and handle that definition in the simulation runner as desired) 

Either way that old way does not work for that...

PR #482, opened 2026-08-03
Update heat exchangers and refactor weighted sensor calculation

hmvp:


zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -131,45 +132,49 @@ class ThrustersSensorValues(ThrsValues):
...
+                    self.thrusters_switch_aft.position_rel,
+                    self.thrusters_switch_fwd.position_rel,
+                ],
+                [self.thrusters_flow_aft, self.thrusters_flow_fwd],
+                Stamped[LMin].stamp(0.0),
             )
         )
 
+        return sensor.CalculatedFlow.from_weighted_sensors(
+            [self.thrusters_mix_exchanger.position_rel],
```
hmvp: Position should be inverterd (1-pos)

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -131,45 +132,49 @@ class ThrustersSensorValues(ThrsValues):
     )
     @property
     def thrusters_temperature_recovery(self) -> sensor.CalculatedTemperature:
-        total_flow = (
-            self.thrusters_flow_aft.flow.value + self.thrusters_flow_fwd.flow.value
+        return sensor.CalculatedTemperature.from_weighted_sensors(
+            [self.thrusters_flow_aft.flow, self.thrusters_flow_fwd.flow],
+            [self.thrusters_temperature_aft, self.thrusters_temperature_fwd],
+            self.thrusters_temperature_aft.temperature,
         )
```
hmvp, line 136: We lost None as response if total flow is zero. I am nu sure if that is desired or not. It is now self.thrusters_temperature_aft.temperature. Which seems to me to be equally as bad/good.

@aljevandam Opinions?

hmvp (commented):
Some remarks

zero-thrs-control/src/thrs/input_output/definitions/sensor.py
```diff
@@ -55,10 +55,44 @@ def from_max_temperature(cls, sensors: list[TemperatureSensor]):
             )
         )
 
+    @classmethod
+    def from_weighted_sensors(
+        cls,
+        ratios: Sequence[Stamped[Ratio | LMin]],
+        sensors: Sequence[TemperatureSensor],
+        default_if_all_closed: Stamped[Celsius],
```
aljevandam: default_if_all_closed doesn't make sense for weighted sensors right? Shouldn't it be default_if_no_weights are something like that (i.e. if sum(ratios) is not 1) 

Also, it is a bit wonky to call a list of flow the 'ratios' - rather call the weights and convert them to ratios inside the function if needed? 

Something like if ratios = weights / sum(weights) 

if sum(ratios) = 0 return default

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
aljevandam: I'd say return None (and show nothing in the mimic) if the calculation can't be made. Then it's most clear what's going on (showing a random number would be misleading right?)

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -131,45 +132,49 @@ class ThrustersSensorValues(ThrsValues):
     )
     @property
     def thrusters_temperature_recovery(self) -> sensor.CalculatedTemperature:
-        total_flow = (
-            self.thrusters_flow_aft.flow.value + self.thrusters_flow_fwd.flow.value
+        return sensor.CalculatedTemperature.from_weighted_sensors(
+            [self.thrusters_flow_aft.flow, self.thrusters_flow_fwd.flow],
+            [self.thrusters_temperature_aft, self.thrusters_temperature_fwd],
+            self.thrusters_temperature_aft.temperature,
```
aljevandam: set default to None

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -131,45 +132,49 @@ class ThrustersSensorValues(ThrsValues):
     )
     @property
     def thrusters_temperature_recovery(self) -> sensor.CalculatedTemperature:
```
aljevandam, line 130: perhaps thrusters_temperature_return is more accurate

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -131,45 +132,49 @@ class ThrustersSensorValues(ThrsValues):
...
+        json_schema_extra=computed_meta(
+            component_type="calculated_flow", included_in_fmu=False
+        )
+    )
+    @property
+    def thrusters_seawater_exchanger_flow(self) -> sensor.CalculatedFlow:
+        total_flow_from_thrusters_past_exchanger_valve = (
+            sensor.CalculatedFlow.from_weighted_sensors(
+                [
+                    self.thrusters_switch_aft.position_rel,
```
aljevandam: Using the valve positions as weights won't work I think (esp for switch valves) as the flow is not linear with valve position. Also, we want total flow here (not weighted average?) 

Isn't the flow past the mix_exchanger given by flow_aft + flow_fwd (always?) 

It seems that inferring the flow past the exchanger and the Tin and Tout of the exchanger is not really possible here..BUT we can get the dT of exchanger including the mix valves here as (we probably want to tweak the design a bit here... ) 

temperture_supply = 
if switch_recovery is closed: 
             if switch_aft and switch_fwd to A:
                          return thrusters_temperature_return 
             elif switch_aft to B and switch_fwd to A:  temperature_aft
             elif switch_aft to A and switch_fwd to B:  temperature_fwd
             else: return None 
if switch_recovery is open and switch_aft to B and switch_fwd to B:
           return temperature_recovery_mix 
else: return None 

temperature_return =  thrusters_temperature_supply

flow = flow_aft + flow+fwd
       
does that make sense?

zero-thrs-control/src/thrs/input_output/definitions/sensor.py
hmvp: Yeah weights is a better term indeed

zero-thrs-control/src/thrs/input_output/definitions/sensor.py
```diff
@@ -55,10 +56,55 @@ def from_max_temperature(cls, sensors: list[TemperatureSensor]):
...
 
 class CalculatedFlow(ThrsValues):
     flow: Stamped[LMin]
 
+    @classmethod
+    def from_weighted_sensors(
+        cls,
+        weights: Sequence[Stamped[Ratio | LMin]],
+        sensors: Sequence[FlowSensor | Self],
+        default_if_all_closed: LMin,
```
aljevandam, line 85: closed is a bit strange here, as it implies the weights are always valve ratios. 

default_if_weights_zero or something would make more sense

zero-thrs-control/src/thrs/input_output/definitions/sensor.py
```diff
@@ -123,6 +177,50 @@ def valves_open_closed(open_valves: list[Valve], closed_valves: list[Valve]) ->
...
 
 
+def weighted_combined_measurement[
+    Measurement: Celsius | LMin,
+    Default: Celsius | LMin | None,
+](
+    weights: Sequence[Stamped[LMin | Ratio]],
+    measurements: Sequence[Stamped[Measurement]],
+    default_if_zero_weight: Default,
+) -> Stamped[Default]:
```
aljevandam, line 187: weird generics. I was expecting        Default: Measurement | None

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -124,45 +128,91 @@ class ThrustersSensorValues(ThrsValues):
     )
     @property
     def thrusters_temperature_recovery(self) -> sensor.CalculatedTemperature:
-        total_flow = (
-            self.thrusters_flow_aft.flow.value + self.thrusters_flow_fwd.flow.value
+        """The average temperature after thrusters before optional mix with pcm."""
```
aljevandam, line 131: yeah so I'd call this the return thrusters_temperature_return..

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -124,45 +128,91 @@ class ThrustersSensorValues(ThrsValues):
...
+
+    @computed_field(
+        json_schema_extra=computed_meta(
+            component_type="calculated_temperature", included_in_fmu=False
+        )
+    )
+    @property
+    def thrusters_temperature_return(self) -> sensor.CalculatedTemperature:
+        """
+        The temperature at after the thrusters and mix with pcm.
```
aljevandam, line 146: The temperature after the thrusters and mix with pcm

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -124,45 +128,91 @@ class ThrustersSensorValues(ThrsValues):
...
+
+        If this was an actual sensor it would have been located just before the inlet to the seawater exchanger.
+        """
+        if sensor.valves_open_closed(
```
aljevandam, line 150: so this it the case 'if cooling_mode' (switches to A) then return thrusters_temperature_return (can perhaps add a a comment as below as note to self - I'm sure that will be useful in the future

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -124,45 +128,91 @@ class ThrustersSensorValues(ThrsValues):
...
-            and self.thrusters_temperature_aft.temperature.value is not None
-            and self.thrusters_temperature_fwd.temperature.value is not None
+
+    @computed_field(
+        json_schema_extra=computed_meta(
+            component_type="calculated_temperature", included_in_fmu=False
+        )
+    )
+    @property
+    def thrusters_temperature_return(self) -> sensor.CalculatedTemperature:
```
aljevandam, line 144: Call this thrusters_temperature_pre_cooler or something? (it is the return temperature mixed with pcm supply right)

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -124,45 +128,91 @@ class ThrustersSensorValues(ThrsValues):
...
         json_schema_extra=computed_meta(
             yard_tag="50001001", component_type="heat_exchanger", included_in_fmu=False
         )
     )
     @property
     def thrusters_seawater_exchanger(self) -> sensor.HeatExchanger:
-        return sensor.HeatExchanger(
-            delta_t=Stamped.stamp(0), heat=Stamped.stamp(0)
-        )  # TODO: Find where this should come from
+        temperature_supply = self.thrusters_temperature_return.temperature
```
aljevandam, line 192: thrusters_temperature_pre_cooler

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
```diff
@@ -124,45 +128,91 @@ class ThrustersSensorValues(ThrsValues):
...
+        temperature_return = self.thrusters_temperature_supply.temperature
+        flow = self.thrusters_flow.flow
+
+        # DeltaT is slightly more difficult since we need to account for the part that does not flow past the exchanger
+        delta_t = Stamped.combine(
```
aljevandam, line 197: I get to, assuming a linear relation between flow an valve position:

delta_t  = (1 - thrusters_mix_exchanger.position_rel)*(temperature_return.value - temperature_supply.value) 

and my own personal opinion is to leave the delta_t out (but keep it in for now.. ) 

Regardless, I'd implement heat as 

heat = (temperature_return.value - temperature_supply.value)*flow*WATER_HEAT_TRANSER_CONVERSION (not using delta_t) 

so no hiden assumptions needed in there.

zero-thrs-control/src/thrs/input_output/modules/thrusters.py
hmvp: Can't rename, needed in fmu

zero-thrs-control/src/thrs/input_output/definitions/sensor.py
```diff
@@ -123,6 +177,50 @@ def valves_open_closed(open_valves: list[Valve], closed_valves: list[Valve]) ->
     )
 
 
+def weighted_combined_measurement[
+    Measurement: Celsius | LMin,
+    Default: Celsius | LMin | None,
```
Boudewijn26, line 182: Shouldn't this just be one generic and then have the type of `default_if_zero_weight` be `Measurement | None`

Ah, the return type should only contain `None` if the default contains `None`. I think this qualifies for an `@overload` instead of an extra generic

zero-thrs-control/src/thrs/input_output/definitions/sensor.py
Boudewijn26: Can't do that in Python unfortunately (generics can't be bounded by generics)

PR #484, opened 2026-08-03
Prefix Thrusters module controller names

hmvp:


zero-thrs-control/src/thrs/input_output/definitions/sensor.py
aljevandam: not part of this pr?

aljevandam (commented):
Looks like there's some changes in here that should ve part of refactor-weighted-calcs?

zero-thrs-control/src/thrs/input_output/definitions/sensor.py
hmvp: No that is the middle pr of this stack See https://github.com/foundation-zero/zero/pull/482

aljevandam (approved):

PR #488, opened 2026-08-04
Update ref values mapping

aljevandam:


aljevandam:
This should give more sensible reference value results. We still need to figure out how to get a sensible percentage out of the trim values for the relative positions

Boudewijn26 (approved):

